PR #470, opened 2013-12-04
Batarray

c-cube:
two functions for BatArray, regarding sorted arrays.


gasche:
Split was merged in master, will be included in 3.0.

Issue #109, opened 2011-02-04
BatIO.input_channel should have cleanup param default to true


mdekstrand:
If you need `cleanup:false` for `BatIO.input_channel`/`BatInnerIO.input_channel`, you're proably doing something wrong or at least very strange.  The default should be `true` so that IO naturally cleans things up.

This is a breaking change, so it needs to wait for 2.0.

gasche:
Fixed by merging #671.

PR #489, opened 2013-12-30
small patch for #109 , with cleanup of BatIO.input enabled by default

c-cube:
couldn't comment on the issue (too old!? github bugs), so here's a patch just in case


gasche:
Merged in master, should be in 3.0.

Issue #10, opened 2009-11-20
LWT support


thelema:
We should provide monadic interfaces to our data structures so LWT can interface with them better

PR #501, opened 2014-01-07
Monads everywhere

c-cube:
Ok, now this is a bigger PR than usual. I started it after I looked at #10, for monads like `'a Lwt.t`. It adds monad instances to `batSeq`, `batList` and `batLazyList`, adds a `BatInterfaces.Traversable` module signature that looks like

``` ocaml
module type Traversable = sig
  type 'a t  (* container *)

  type 'a m (* monad *)

  val mapM : ('a -> 'b m) -> 'a t -> 'b t m

  val foldM : ('a -> 'b -> 'a m) -> 'a -> 'b t -> 'a m

  val sequence : 'a m t -> 'a t m
end
```

and functors `Traversable` that take a `BatInterfaces.Monad` and return a traversable for this monad and container, to aforementioned modules as well as `BatEnum`. For `BatEnum`, a similar `WithMonad` functor was already present so I modified it, but didn't rename it to preserve backward compatibility.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,7 @@ PREPROCESSED_FILES = \
  src/batBuffer.ml \
  src/batBytes.ml src/batBytes.mli \
  src/batDigest.ml \
+ src/batFormat.ml \
  src/batHashtbl.ml \
  src/batInnerPervasives.ml \
  src/batList.ml \
--- a/src/batEnum.ml
+++ b/src/batEnum.ml
@@ -1434,40 +1434,40 @@ end
 
 module WithMonad (Mon : BatInterfaces.Monad) =
 struct
+  type 'a t = 'a mappable
   type 'a m = 'a Mon.m
 
-  let sequence enum =
-    let (>>=) = Mon.bind and return = Mon.return in
-    (* We use a list as an accumulator for the result sequence
-       computed under the monad. A previous version of this code used
-       a Queue instead, which was problematic for backtracking
-       monads. Due to the destructive nature of Enums, the current
-       version will still be problematic but at least the result will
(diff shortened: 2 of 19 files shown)


gasche:
For the record, I just started looking at this PR again, writing tests to eventually be done with it. I'm in the List Traversable instance, and I found a small bug:

```ocaml
let rec map_m f l = match l with
| [] -> M.return []
| x::l' ->
  map_m f l' >>= fun l' ->
  f x >>= fun x ->
  M.return (x::l')
```

This will perform the side-effects right-to-left instead of the expected left-to-right (be it only to match with `List.map`'s semantics). It needs to be rewritten as (I did this in my local branch):

```ocaml
  f x >>= fun x ->
  map_m f l' >>= fun l' ->
  M.return (x::l')
```

(and then probably stick an accumulator and `List.rev` to make this tailrec)

gasche:
[gasche/monads](https://github.com/gasche/batteries-included/tree/monads)

gasche:
@c-cube , could you force-push the content of my [monads](https://github.com/gasche/batteries-included/commits/monads) branch (which includes your patches) into your own branch, so that we can keep the commits and discussion in this PR?

UnixJunkie:
this PR is one of the cool stuffs we have not merged yet

UnixJunkie:
One day, we should really do a major release...

UnixJunkie:
This is an interesting old PR.  We should revive it.
I wonder if it is impacted by the new syntax for monads that was added recently.
I never use monads in OCaml, so I am not very aware of all that.

UnixJunkie:
closing because too old without any action

Issue #492, opened 2014-01-03
hash functions in modules


c-cube:
having a sensible hash function in batInt, batString, etc. would make many functor applications easier (hashtbl and weak tables, for instance). Also, it can help make hashing specific to types:

``` ocaml
let hash (i : int) = i land max_int
let hash (s : string) = (* function that hashes a string, or just Hashtbl.hash *)

(* ... *)

let hash (b : Big_int.big_int) = String.hash (Big_int.string_of_big_int b)
```

A **very** useful thing to have, also, is hash combinators ([example](https://github.com/c-cube/logtk/blob/master/src/hash.ml)) for lists, arrays, enums, similarly to printing functions.

PR #507, opened 2014-01-13
Hash functions

c-cube:
Early draft for #492, as simple as possible (and even more). I just implemented hash functions for a few modules, but it should be easy to do so for other structures (e.g. `BatLazyList` or `Tuple2`, `Tuple3`...).

### DIFF
--- a/src/batArray.ml
+++ b/src/batArray.ml
@@ -513,6 +513,13 @@ let compare cmp a b =
    compare (fun x y -> -(Pervasives.compare x y)) [|2;1|] [|1;2|] = -1
 *)
 
+let hash h_elem a =
+  let h = ref 0 in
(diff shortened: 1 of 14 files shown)


UnixJunkie:
Is this really needed?
I am not against it but I am not a specialist of hashing either.

UnixJunkie:
We should do a vote, or we should go vote, depending on where you live.

UnixJunkie:
I am not qualified to review this PR: if I don't use a Hashtbl.t, then I don't hash anything.
Note however that this PR was submitted by a very skilled hacker.
We should probably merge it.

c-cube:
probably no need for that, unless someone shows up here and asks for the feature :slightly_smiling_face:

PR #512, opened 2014-01-19
high level file locking file interface

c-cube:
Not sure if useful enough for batteries, but I found this code nice for coarse-grained concurrent updates on files.

### DIFF
--- a/src/batUnix.ml
+++ b/src/batUnix.ml
@@ -175,3 +175,22 @@ let is_directory fn = (lstat fn).st_kind = S_DIR
 let rec restart_on_EINTR f x =
   try f x
   with Unix_error(EINTR, _, _) -> restart_on_EINTR f x
+
+(** {6 Locking} *)
+
+let with_file_lock ~kind filename f =
+  let lock_file = openfile filename [O_CREAT; O_WRONLY] 0o644 in
+  let lock_action = match kind with
+    | `Read -> Unix.F_RLOCK
+    | `Write -> Unix.F_LOCK
+  in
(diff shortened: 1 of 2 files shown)


UnixJunkie:
I am not sure we need this in batteries.
There are libraries providing semaphores (ocamlnet and posix-semaphore in opam) and files can be opened in append mode...
I don't remember having locked a file in my whole programmer life.
I think this PR should be closed, unless others feel this is needed/useful.

rixed:
I, for one, often use something along the lines of the proposed functions.

I have a couple of remarks though:
First, the call to the user function must be protected by a finally to make sure the function is never exited with the lock still on. Also, I'd like this user function to be passed the file descriptor (maybe as a BatIO channel for consistance, although I tried to avoid them as much as possible for being unpredictable). If the file descriptor is passed to the user function in one form or another, though, then of course one has to lseek back to the beginning of the file before unlocking.

I can offer a modified patch if anyone's interested and @c-cube is not interested in contributing it any longer, which after such a long time would be understandable :)

UnixJunkie:
Do not hesitate to send your patch (this PR has rather positive feedback from three vetted OCaml programmers, I am the only dissident voice).
I would prefer avoiding IO (I personally avoid it for performance reasons, but also one day
it might get out of batteries, or many modules made so as to not depend on it anymore
so that they can be shipped separately).

rixed:
Proposed new PR: https://github.com/ocaml-batteries-team/batteries-included/pull/904

UnixJunkie:
there is a superseeding PR

Issue #503, opened 2014-01-07
combination/permutations combinators


c-cube:
Someone on IRC asked whether some OCaml library provided functions to generate permutations or combinations of a bunch of elements. I didn't see any in Batteries (although I may have missed them in some module); they seem very useful for many algorithms and are not trivial to implement, so I think it may be a good idea to add at least `BatList.combinations` and `BatList.permutations`. An in-place generator for arrays is also doable (see C++'s STL).

gasche:
Where are we on this? @c-cube recalled us that this PR is undecided (thanks!), but it's not clear to me what the best move is. The code as proposed by @Chimrod is not mergeable as-is (because of the french wordings), and there is a debate on the interface.

I wouldn't mind adding a `BatCombinatorics` module with a more polished version of @Chimrod code, but given the absence of dependencies on anything else in Batteries I also think that it would be served just as well as an independent package available on opam, that non-Batteries users may have less qualms including. (There already is [backtracing/combine](https://github.com/backtracking/combine), but it seems more about constraint resolution than arithmetic enumerations.)

c-cube:
Side note: there is also

```ocaml
val Gen.permutations : 'a gen -> 'a list gen                                                 
val Gen.permutations_heap : 'a gen -> 'a array gen 
val Gen.combinations : int -> 'a gen -> 'a list gen
```

UnixJunkie:
Now we have those in BatLazyList, thanks to @c-cube

PR #518, opened 2014-01-21
Added a combination generator for Enum

Chimrod:
Implement an enumeration over combination #503

This an implementation of mathematical Combination ( https://en.wikipedia.org/wiki/Combination ), wich work with or without repetitions.

### DIFF
--- a/src/batEnum.ml
+++ b/src/batEnum.ml
@@ -1158,6 +1158,72 @@ let slazy f =
 
 let delay = slazy
 
+let combination ?(repeat=false) n k =
+  let binomial n p =
+    let binom n p =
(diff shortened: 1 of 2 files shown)


UnixJunkie:
The ocamldoc should have some examples, so that typical use cases can be understood.

UnixJunkie:
And, I'll leave it to fans of the Enum module to decide if new things must be added in it.

rixed:
I vote for it.

rixed:
5 years later I think this is still desirable. Are there objections for merging this?

UnixJunkie:
The corresp. ChangeLog entry is missing.
Appart from this, it looks OK to merge.

PR #530, opened 2014-02-04
correct the following compilation warnings:

UnixJunkie:
Warning: tag "package" does not expect a parameter, but is used with parameter "oUnit"
Warning: tag "package" does not expect a parameter, but is used with parameter "QTest2Lib"

PS: we have 11 _tags file in the batteries tree, amazing...

### DIFF
--- a/qtest/_tags
+++ b/qtest/_tags
@@ -1,3 +1,3 @@
 true: threads, debug
-<all_tests.*>: package(oUnit), package(QTest2Lib)
+<all_tests.*>: oUnit, QTest2Lib
 


UnixJunkie:
@c-cube do you manage to remove those warnings?

gasche:
I still think that passing `-use-ocamlfind` in the right place(s) in the Makefile should solve these warnings without touching those tags, which look fine. I wouldn't support removing the tags just to fix warnings.

UnixJunkie:
I don't see those warnings anymore.

PR #545, opened 2014-03-06
BatSeq: Fix specification of `filter`

nbraud:
The previous specification for `filter` (and other functions) was nonsensical as it considered elements of a sequence that “occur” after an infinite number of other elements.

For instance, it could have lead one to understand that `filter (fun x -> x == 0)` should terminate on the sequence `1 1 1 1 ...`.


UnixJunkie:
I don't thing the current documentation of BatSeq.filter is very good/clear.
And, this PR doesn't make it significantly better.

PR #547, opened 2014-03-08
Add enum/of_enum and implode/explode to BatUTF8

pminten:
I'm not quite sure if UTF-8 in these source files (in the test comments) is good style.


UnixJunkie:
Consensus was not reached.

PR #554, opened 2014-03-26
Batteries light: proposals for the next major version

c-cube:
This PR should probably be merged in a new branch (for v3?). It contains three major changes from the current 2.2 version:
- Batteries is split into several libraries built with oasis (the `configure` file allows to choose which ones are compiled and installed). The main one (called `batteries`) doesn't depend on `Unix`, `Bigarray` nor `Num`. `BatPervasives` is moved into a sub-library because it provides some IO primitives that depend on `Unix`; to retrieve the usual behavior (`open Batteries`) the library `batteries.full` has to be used.
- `Enum` is replaced by the structural type `'a Gen.t = unit -> 'a option`, and a `BatGen` module is provided to deal with the new iterators
- Some experimental support is provided for serialization with `BatConv`.


UnixJunkie:
Nice try, but should use dune, not oasis.

PR #557, opened 2014-04-04
Balanced binary trees: enumeration starting from an arbitrary key

gasche:
``` ocaml
val enum_from : key -> 'a t -> (key * 'a) BatEnum.t
(** [enum_from k t] return an enumeration as [enum], but starting from
    the first key equal or greater than [k] *)


val backwards_from : key -> 'a t -> (key * 'a) BatEnum.t
(** [backwards_from k t] return an enumeration as [backward], but
     starting from the last key smaller or equal to [k] *)
```

Of course the performance of this are only as good than our enumeration library, but the code does not depend on it (Map and Set already had manually-derived control-inversed iterators that make these kind of things convenient) so it will be trivial to change to, say, `Sequence` if we want to in the future.

### DIFF
--- a/src/batMap.ml
+++ b/src/batMap.ml
@@ -255,8 +255,6 @@ module Concrete = struct
       else
         let (lr, pres, rr) = split key cmp r in (join l x d lr, pres, rr)
 
-  type ('key,'a) iter = E | C of 'key * 'a * ('key,'a) map * ('key,'a) iter
-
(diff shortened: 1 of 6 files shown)


UnixJunkie:
It looks like this needs more work in order to be merged.

UnixJunkie:
PR was stalled for so much time that it became a fossil.

PR #571, opened 2014-05-26
add BatSubstring.sub analagous to String.sub

rgrinberg:
makes it easier to replace `String` uses with `BatSubstring`


UnixJunkie:
This should be an issue, not a PR.
I am closing it because I am trying to do one pass over our list of PRs (and don't want to see it
next time I do).
Please reopen as an issue if still interested.

PR #578, opened 2014-07-21
instead of a pair of enums, Enum.combine now takes two enums (it was cur...

UnixJunkie:
...ried)


gasche:
Merged in master, will be in 3.0.

PR #596, opened 2014-10-09
BatUTF8.enum, based on BatString.enum / BatList.enum

unhammer:
Is there a reason for not having this in here? 

I'm new to OCaml, so I probably made some errors here, but it seems useful anyway to have a BatUTF8.enum

### DIFF
--- a/src/batUTF8.ml
+++ b/src/batUTF8.ml
@@ -252,6 +252,17 @@ let enum s =
           c
         end
     )
+(*$T enum
+   "" |> of_latin1 |> enum |> BatList.of_enum = []
+   "foo" |> of_latin1 |> enum |> BatList.of_enum = List.map BatUChar.of_char ['f'; 'o'; 'o']
+   let e = of_latin1 "abcdef" |> enum in \
+   for _i = 0 to 2 do BatEnum.junk e done; \
+   let e2 = BatEnum.clone e in \
+   let to_string en = BatList.of_enum en |> List.map BatUChar.char_of |> BatString.implode in \
+   to_string e = "def" && to_string e2 = "def"
+   init 3 (fun i -> BatUChar.of_int (1211+i)) |> enum |> BatList.of_enum = List.map BatUChar.of_int [1211;1212;1213]
+*)
+(* The last test checks that we can make a round-trip of non-ASCII strings like "һҼҽ" *)
 
 
 let escaped = String.escaped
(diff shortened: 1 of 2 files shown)


gasche:
I'm sorry we failed this pull-request: it has been sitting here for much longer than it should, as neither thelema nor I had much time to contribute recently.

@unhammer , if you are still interested in this contribution, the code looks reasonable and it should be mergeable short-term. The one thing that is currently missing is a test: it was not yet the policy at the time BatUTF8 was contributed, but Batteries only accept new functions when they comes with a test. Adding inline tests is very easy, just reproduce what is done in most modules *other* than BatUtf, eg. in BatString.ml (it is just a magic comment syntax). Then `make qtest` should work.

I think the PR would be in mergeable state when we have basic tests that it works as expected.

I would of course perfectly understand if you had moved to something else since October 2014. In that case, do feel free to say so, we can probably pick up the PR where it stopped.

unhammer:
Some tests – look OK? (first batch just from BatString, then one that checks it doesn't split a wide char)

gasche:
I think this looks OK. Do you know if it would be possible to write the last test in a way that (is still readable but) would be robust to a change of encoding of the source file? If there is no easy way to do that, I would merge the patch as-is.

(I'm a bit surprised that `BatUTF8`, which I rarely use myself, only has `of_latin1` as non-unsafe import function, as it seems odd to call it on a literal string written in a utf8-encoded file, and I would expect something like `of_utf8 : string -> t` or at least a less scary name than `of_string_unsafe`, but then that is the design of the library as it is now and not for this PR to fix.)

unhammer:
Yeah, I was wondering myself if it was OK to include the literal (although the file does have "UTF8" in its name heh). I guess something like

    init 3 (fun i -> BatUChar.of_int (230+i)) |> enum|>BatList.of_enum = List.map BatUChar.of_int [230;231;232];;

could work?

gasche:
To a non-UTF8-expert such as me, it is not as clear that those characters are wide chars; I think just a mention in comments of what they actually look like (it's fine if we have utf8-only comments, I think) would help. In any case, I think that's probably a better test -- feel free to update the PR.

unhammer:
it didn't seem like nested comments worked, so I put the comment below, is that ok?

gasche:
This is fine. Would you rebase and squash the first three (implementation-related) commits together (no need to expose the intermediary inefficient state), and the last three test-related commits together? `push -f` works in silently updating the pull request.

unhammer:
there

gasche:
Thanks a lot, merged!

PR #602, opened 2015-01-03
Add some missing functions from Format

choeger:
These useful helper-functions were introduced in 4.02 but are missing in Batteries.


gasche:
If this wasn't caught by `batteries_compattest.ml`, we need to update it as well.

choeger:
See also:

https://github.com/ocaml-batteries-team/batteries-included/blob/master/src/batteries_compattest.ml#L23

Issue #560, opened 2014-04-24
BatHashtbl modify functions never resize


jj-issuu:
The functions `modify_opt` and `modify_def` in BatHashtbl never resize the hash table, so they have O(n) performance if the table is never modified through a function like `add` or `replace`. This should at the very least be documented, but resizing really must be implemented in order for those functions to be useful.

Looking at the source code, it's hard to see how to keep the non-inserting case highly efficient while also supporting resize and not depending too much on implementation details of stdlib/hashtbl.ml.

Perhaps the most robust solution is to implement  `modify_opt` and `modify_def` through the standard add/remove/replace/find functions. This makes the non-inserting cases slower but keeps all cases at O(1) average.

UnixJunkie:
I think this is a performance bug.
But maybe it was corrected?
I see several merges related to this.

jj-issuu:
I think the changes in #603 optimized this code so much that it can't be considered a performance bug. There's still room for improvement, but that would come at a high cost in code complexity and dependence on how Hashtbl is implemented in the standard library.
IMHO, this issue can be closed.

UnixJunkie:
Thanks for the feedback.

PR #603, opened 2015-01-15
Optimize hashtbl modify_{def, opt} functions

andersfugmann:
This pull request optimizes the Hashtbl.modify functions to eliminate an extra lookup in the case the key is already in the hash table.

Hashtbl.modify_opt still does an extra lookup when adding the key to the hash table in order to make sure the hash table is grown if needed (by call to Hashtbl.add).

The pull request also contains a change to Hashtbl.modify, to use the new key instead of the old key to follow the semantics of Hashtbl.replace.
This change might put more pressure on memory as the new key will to be copied to the major heap if not already there (if the hashtbl is on the major heap). OTOH, the change follows the semantics of Hahtbl.replace and makes it easier to control which keys are referenced or not (Consider the case if the key is referenced from the value itself, e.g. Hashtbl.modify k (k, v) h. or if a custom equal function is used).

### DIFF
--- a/src/batHashtbl.mlv
+++ b/src/batHashtbl.mlv
@@ -188,7 +188,11 @@ let modify_opt key f h =
   let hc = h_conv h in
 
   let rec loop = function
-    | Empty -> raise Hashtbl_key_not_found
+    | Empty ->
+      (* Inserting an element might require a resize of the hash table.
+         We rely on Hashtbl.add function to grow the hashtbl if needed
+         instead of duplicating logic from the OCaml standard library. *)
+      raise Hashtbl_key_not_found
     | Cons(k,v,next) ->
       if k = key then
         match f (Some v) with
@@ -206,7 +210,10 @@ let modify_opt key f h =
   | Hashtbl_key_not_found ->
     begin match f None with
       | None -> ()
-      | Some v -> add h key v
+      | Some v ->
+        (* Add the element to make sure the hashtbl is grown correctly if
+           needed. *)
+        add h key v
     end
 
 (*$T modify_opt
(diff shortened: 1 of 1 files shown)


gasche:
I haven't looked at the change in detail yet (I won't have time before next week), but I would like to see careful comments about how this new implementation avoids bug  #560 . I agree with the proposed `Hashtbl.modify` change.

andersfugmann:
I have added comments to the implementation on how the table (buckets) are grown when adding elements though calls to modify_opt..

src/batHashtbl.mlv
```diff
@@ -183,16 +183,38 @@ let length h = (h_conv h).size
...
-     (match f None with
-      | Some v -> add h key v
-      | None -> ())
+  let hc = h_conv h in
+
+  let rec loop = function
+    | Empty ->
+      (* Inserting an element might require a resize of the hash table.
+         We rely on Hashtbl.add function to grow the hashtbl if needed
+         instead of duplicating logic from the OCaml standard library. *)
```
gasche: Sorry for the nitpicking, but this comment should be above the `| Empty` line, not under it (as placed it seems to comment on the `raise Not_found ...` line below, or in any case be specific to the `Empty` case. If you are uneasy with separating the head `function` and its first case with a comment, feel free to have the comment first, and `function` second on its own line (the `|` does not need to be indented wrt. `function`).

andersfugmann:
Ok. I've moved the comment above the Empty case.

gasche:
Could we do without the exceptions? I find the control-flow in the original code (which can be seen as the red parts of the diff in ec3d12062a544a , the naive fix to #560 ) easier to follow. Exception-raising-and-catching raises efficiency issues on some exotic backends (eg. js_of_ocaml) so I'd rather only use them when they're a clear win.

andersfugmann:
I agree that exceptions as flow control should be avoided, however I dont really see an obvious
clean way to break the flow and call add without updating the bucket (the call to Array.unsafe_set).

The old implementation always updates the bucket, hence everything can be handled directly in the loop, which makes the flow easier to read.

To get rid of exceptions, the loop function needs to be change to e.g. an option type, but then the inner loop cannot be implemented as a simple recursive function. 

Do you have an dies on how to implement in a clean way without exceptions. A local ref 'call_add' to indicate if the result from the inner loop should be used or if Hashtbl.add should be called? - But that would be a marginally more inefficient on non-exotic platforms than unrolling the stack right away (a new unused list of bucket would still need to be created only to be discarded).

gasche:
I was a bit too optimistic about this change. You're entirely correct that given the current non-tail-recursive implementation of bucket list lookup, it's hard to express the same control flow without the exceptional escape.

Issue #604, opened 2015-01-22
Chaining operators for 'a option and 'a result


herry13:
Hi All, thanks for this great software. Question: is there any operator for chaining functions which return 'a option / 'a result? If not, I want to propose my implementations to be added into batteries which I found very useful and simple to be understood. The signatures of the functions are:

`val (|?>) : 'a option -> ('a -> 'a option) -> 'a option`

`val fold_opt : ('a -> 'b -> 'b option) -> 'a list -> 'b -> 'b option`

`val (||>) : ('a, exn) result -> ('a -> ('a, exn) result) -> ('a, exn) result`

`val fold_res : ('a -> 'b -> ('b, exn) result) -> 'a list -> 'b -> ('b, exn) result`

hcarty:
For `(|?>)` we already have `BatOption.bind`.  For `(||>)` we have `BatResult.Monad.bind`.  For consistency they should both be spelled `(>>=)` which `result` already has as `BatResult.Infix.(>>=)`.  Adding `BatOption.Infix.(>>=)` seems like a reasonable thing to do.  `BatResult.Infix.(|?)` may make sense too.

I don't have an opinion one way or the other on the `fold` functions, other than that they should probably live in `BatList`.

herry13:
@hcarty thanks for the pointers. I prefer to use different infix symbols so that a module can deal with `option` and `result`.

I agree for `fold` functions, they should be in `BatList`.

hcarty:
For what it's worth, local module opens give you some flexibility with overlapping operator names.  That said, depending on the structure of your code they can make things pretty noisy.

herry13:
@hcarty I understand that name consistency is important. However, my current project has functions that have to deal with both `option` and `result`. So overlapping operator names won't work in my case.

gasche:
The right way to do those folds is to implement a folding function over any monad. @c-cube has implemented monad-generic `fold` (and `map` and `traverse`) for many sequence libraries (including lists) in the pull request #501 , but the code is sitting outside unmerged because @c-cube had no time to polish the code to meet the merge requirements (what it lacks are just unit tests). If anyone here is willing to implement the necessary unit tests, that would be a warmly appreciated contribution.

herry13:
@gasche thanks for the comment. I tried @c-cube codes for `option` and it works:

```ocaml
module TOption = Traverse(BatOption.Monad)
let fold_opt = TOption.foldM
```

[...]

PR #605, opened 2015-01-23
Implement BatOption.{Monad,Infix}.(>>=)

herry13:
Implement `BatOption.Monad.(>>=)` and `BatOption.Infix.(>>=)` as discussed in #604.

### DIFF
--- a/src/batOption.ml
+++ b/src/batOption.ml
@@ -193,8 +193,6 @@ struct
   let bind m f = match m with
     | None -> None
     | Some x -> f x
-
-  let (>>=) = bind
 end
 
 let bind = Monad.bind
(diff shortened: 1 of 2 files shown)


gasche:
The way we have done things so far is rather to have the infix operator only in the `Infix` module (for example `?|` is not defined outside `Infix`). If you propose a PR with just `Infix.(>>=)` we can merge it immediately. If you/hcarty think also having `Monad.(>>=)` is important, we can discuss it.

herry13:
Honestly, I don't use use `Monad.(>>=)`. I added it only for consistency between `BatOption.Monad` and `BatResult.Monad`.

herry13:
@gasche I had removed `BatOption.Monad.(>>=)`. Could you please merge it? Thanks.

gasche:
I merged (a squashed version of) this PR, thanks!

PR #610, opened 2015-04-16
added an interleave utility function to the BatEnum module

azarens:
keeps track of transitions in an array of indexes

never polls the expired enum twice

### DIFF
--- a/src/batEnum.ml
+++ b/src/batEnum.ml
@@ -1253,6 +1253,49 @@ let merge test a b =
     if *)
 
 
+let interleave enums =
+  let enums_len = Array.length enums in
+  if not (enums_len > 0) then empty () else begin
+    let available = Array.make enums_len true
+    and next_idx = Array.init enums_len ((+) 1) in
+    next_idx.((Array.length next_idx) - 1) <- 0 ;
+    let rec next_elem idx =
+      match get enums.(idx) with
+       | Some x -> x , next_idx.(idx)
+       | None -> begin
+           available.(idx) <- false ;
+           let rec loop k =
+             let l = next_idx.(k) in
+             if l = idx then raise No_more_elements else
+             if available.(l) then (next_idx.(idx) <- l ; next_elem l) else loop l
+           in loop idx
+         end
+    in
+    from_loop 0 next_elem
+  end
+
+(*$T
+  let e1 = List.enum [ 8 ; 2 ; 5 ; 2 ] and e2 = List.enum [ -5 ; -7 ; -6 ; 2 ; 1 ; -9 ; 2 ] in \
+  let e = interleave [| e1 ; e2 |] in \
+  List.of_enum e = [ 8 ; -5 ; 2 ; -7 ; 5 ; -6 ; 2 ; 2 ; 1 ; -9 ; 2 ]
+*)
+
+(*
+  let let e1 = Enum.empty \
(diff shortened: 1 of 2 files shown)


src/batEnum.ml
```diff
@@ -1237,6 +1237,27 @@ let merge test a b =
...
+             let l = next_idx.(k) in
+             if l = idx then raise No_more_elements else
+             if available.(l) then (next_idx.(idx) <- l ; next_elem l) else loop l
+           in loop idx
+         end
+    in
+    from_loop 0 next_elem
+  end
+
 (*$T
```
gasche: the test below is an inline test for the function `merge` above. You should not separate `merge` from its test (but put your implementation below the test), and also you should provide new tests for the function which did not exist before (see other `(*$T ... *)` blocks for inspiration). Inline tests can be tested with `make qtest` or `TESTABLE=src/batEnum.ml make qtest` to test only the `Enum` module.

azarens: Ok, thank you; will work on that.

thelema:
Also, please add some tests for this new function; simple tests in the style of other enum self tests will do.

thelema:
I like the use of 'next_idx' to efficiently interleave many enum of different length.

azarens:
Done.

src/batEnum.ml
```diff
@@ -1253,6 +1253,49 @@ let merge test a b =
...
+    from_loop 0 next_elem
+  end
+
+(*$T
+  let e1 = List.enum [ 8 ; 2 ; 5 ; 2 ] and e2 = List.enum [ -5 ; -7 ; -6 ; 2 ; 1 ; -9 ; 2 ] in \
+  let e = interleave [| e1 ; e2 |] in \
+  List.of_enum e = [ 8 ; -5 ; 2 ; -7 ; 5 ; -6 ; 2 ; 2 ; 1 ; -9 ; 2 ]
+*)
+
+(*
```
gasche: You can use `(*$R` for multi-line tests (see `src/batDynArray.ml` for example). In any case you should use some test header, such as `(*$T` as above. You should be able to check that changing a number in one of the tests makes `make qtest` fail.

azarens: Ok; will adjust.

src/batEnum.ml
```diff
@@ -1253,6 +1253,49 @@ let merge test a b =
...
+  and e2 = List.enum [ 8 ; 2 ; 5 ; 2 ] \
+  and e3 = List.enum [ -5 ; -7 ; -6 ; 2 ; 1 ; -9 ; 2 ] in \
+  let e = interleave [| e1; e2 ; e3 |] in \
+  List.of_enum e = [ 8 ; -5 ; 2 ; -7 ; 5 ; -6 ; 2 ; 2 ; 1 ; -9 ; 2 ]
+*)
+
+(*
+  let let e1 = Enum.empty \
+  and e2 = Enum.empty \
+  and e3 = Enum.epmyt \
```
gasche: typo here, `Enum.epmyt`

gasche:
Ping? The patch seems in mergeable test as soon as the inline tests are fixed.

azarens:
I've been a bit busy lately ; will get back to this project either after work today, or by the end of the week.

gasche:
I fixed the tests and merged the fixed commit: ffbdf2bc8ad95ae1d96bddefabfd3086b84022fc

PR #616, opened 2015-06-10
Set.at_exn

UnixJunkie:
If it is accepted, I can propose the version not throwing exceptions on top of it soon.

### DIFF
--- a/src/batSet.ml
+++ b/src/batSet.ml
@@ -823,6 +823,7 @@ let at_exn i s = Concrete.at_exn i s
 (*$T
   at_exn 0 (of_list [1;2]) == 1
   at_exn 1 (of_list [1;2]) == 2
+  at_exn 2 (of_list [1;2;3;4;5]) == 3
   try ignore (at_exn 0 empty); false with Not_found -> true
   try ignore (at_exn (-1) (singleton 1)); false \
   with Invalid_argument _msg -> true
--- a/src/batSet.ml
+++ b/src/batSet.ml
@@ -142,6 +142,29 @@ module Concrete = struct
       Empty -> ()
     | Node(l, v, r, _) -> iter f l; f v; iter f r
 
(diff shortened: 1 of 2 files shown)


src/batSet.ml
```diff
@@ -793,6 +818,18 @@ let remove x s = Concrete.remove Pervasives.compare x s
 
 let iter f s = Concrete.iter f s
 
+let at_exn i s = Concrete.at_exn i s
+
+(*$T
+  at_exn 0 (of_list [1;2]) == 1
+  at_exn 1 (of_list [1;2]) == 2
```
gasche: I think you could have a test with a bit more elements in the set -- for example with 5 elements, picking an element that is neither the first nor the last to avoid edge cases.

src/batSet.mli
```diff
@@ -121,6 +121,12 @@ sig
       The elements of [s] are presented to [f] in increasing order
       with respect to the ordering over the type of the elements. *)
 
+  val at_exn: int -> t -> elt
+  (** [at_exn i s] returns element at rank [i] in [s].
```
gasche: You might want to be a bit more explicit about what you mean by `at rank [i]` -- that it is about the ordering.

UnixJunkie:
All done.

gasche:
I think we want a similar function in `Map` for consistency, and this suggests that the name may be problematic: if we use `Map.at`, I think users would easily confuse it with `find`: you would expect to pass a key rather than an integer.

I think that `at_rank` may thus be a better name for this feature. More precisely, this version would be `at_rank_exn`. What do you think?

UnixJunkie:
I'm fine with renaming: you know the code base more than me, if it's a good name for you it will be fine with me.

UnixJunkie:
The feature was accepted from another PR.

PR #618, opened 2015-06-11
tests, doc and implementation for BatSet.pop_min

UnixJunkie:

PR #619, opened 2015-06-11
BatSet.pop max

UnixJunkie:


gasche:
Given that this PR supersedes #618 , maybe we could close #618 and consider this PR the one true place for pop-related changes?

gasche:
The code is in mergeable state.

Could we have `pop_{minx,max}` in `Map` as well? Would it be reasonable as part of this PR, or would you rather merge this now and open a separate PR?

UnixJunkie:
It will be a separate PR.
I track it here so that I don't forget it:
https://github.com/ocaml-batteries-team/batteries-included/issues/622
Maybe next week I will have time for it.

UnixJunkie:
Can this be merged?

PR #621, opened 2015-06-11
BatSet.of_array

UnixJunkie:


gasche:
Have you measured a performance advantage to doing `to_array set` directly rather than `Array.of_enum (to_enum set)`? I'm sure it exists, but having numbers would still be helpful.

gasche:
Thanks for these two solid PRs!

UnixJunkie:
I did not measure. I never use Enum in my code.
If I would really want laziness, I would probably use Stream or LazyList.

PR #623, opened 2015-06-14
Move the prefilterin logic from Makefile to ocamlbuild

### LINKED ISSUE #615: ``make coverage'' error?
Hi, I run into this error when I type ``make coverage'':

```
ocamlbuild -no-links -use-ocamlfind coverage/index.html
Finished, 0 targets (0 cached) in 00:00:00.
File "_tags", line 2, characters 32-42:
Warning: the tag "warn_Ae-29" is not used in any flag declaration, so it will have no effect; it may be a typo. Otherwise use `mark_tag_used` in your myocamlbuild.ml to disable this warning.
+ ocamlfind ocamldep -package bigarray,num,str -syntax camlp4o -package camlp4 -ppopt str.cma -ppopt /Users/bobfang/.opam/system/lib/bisect/bisect_pp.cmo -ppopt -disable -ppopt b -package bytes -modules src/batInnerPervasives.ml > src/batInnerPervasives.ml.depends
I/O error: "src/batInnerPervasives.mlv: No such file or directory"
File "src/batInnerPervasives.ml", line 1:
Error: Error while running external preprocessor
Command line: camlp4 '-I' '/usr/local/lib/ocaml/camlp4' '-parser' 'o' '-parser' 'op' '-printer' 'p' 'str.cma' '/Users/bobfang/.opam/system/lib/bisect/bisect_pp.cmo' '-disable' 'b' 'src/batInnerPervasives.ml' > /var/folders/sg/9m3v78x53_v6qxbnw6kb5dkm0000gn/T/ocamlppb93770

Command exited with code 2.
Compilation unsuccessful after building 34 targets (33 cached) in 00:00:00.
make: *** [coverage/index.html] Error 10
```

Do you know how I can fix it?

gasche:
Fixes #615: "make coverage" is broken.

The current "make coverage" failure occurs as the coverage-checking
tool attempts to access "src/batInnerPervasives.mlv" and does not find
the file. It looks for this file because it is present in the
lexer-directive comments of the preprocessed file
(src/batInnerPervasives.ml), and it makes sense for a coverage
checking to display covered lines in terms of the source file instead
of post-filtering file (which is not what developers see). But because
the prefilter step currently happens purely in Make-land, before
ocamlbuild is invoked, ocamlbuild does not know about this .mlv file
and does not copy it in the _build directory, from where the coverage
preprocessing is started.

The solution to this problem is to teach ocamlbuild about the .mlv
preprocessing step, instead of doing it ahead-of-time in the
Makefile. Now ocamlbuild decides which .mlv to prefilter, and copies
them to _build along the process, which makes "make coverage work
again". This has one downside and some upsides.

Downside:
- The prefiltering logic was previously in the Makefile: (simple, easy
  to read, most programmers are familiar with it. It is now in
  myocamlbuild.ml: less simple, most programmers avoid touching it,
  and it has a bit of cruft due to support for really old ocamlbuild
  versions.

Upsides:
- No more need to explicitly list all files to preprocess in the
  Makefile. This had had been a source of minor inconvenience when
  adding to the set of such files; but admittedly this could be
  automated at the Makefile level as well).

- No more need to use "make clean-prefilter" when unsure whether the
  changes to a .mlv file have been taken into acount; now ocamlbuild
  tracks dependencies for .mlv (and prefilter.byte, which was
  previously mostly untracked by the build system) as any other file,
  and will only rebuild them when needed. This is a major improvement
  in convenience when manipulating these files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,44 +65,22 @@ else
 endif
 endif
 
-PREPROCESSED_FILES = \
- src/batArray.mli src/batArray.ml \
- src/batBigarray.ml src/batBigarray.mli \
- src/batBuffer.ml \
- src/batBytes.ml src/batBytes.mli \
- src/batDigest.ml \
- src/batFormat.ml \
- src/batHashtbl.ml \
- src/batInnerPervasives.ml \
- src/batList.ml \
- src/batMarshal.ml src/batMarshal.mli \
- src/batPervasives.mli \
- src/batPrintexc.mli \
- src/batPrintf.ml src/batPrintf.mli \
- src/batStream.ml \
- src/batSys.mli \
- src/batUnix.ml src/batUnix.mli
-
 .PHONY: all clean doc install uninstall reinstall test qtest qtest-clean camfail camfailunk coverage man
 
-all: prefilter
+all:
 	@echo "Build mode:" $(MODE)
 	$(OCAMLBUILD) $(OCAMLBUILDFLAGS) $(TARGETS)
-
-clean-prefilter:
-	@${RM} $(PREPROCESSED_FILES)
 
 clean:
 	@${RM} src/batteriesConfig.ml src/batUnix.mli batteries.odocl bench.log
 	@${RM} -r man/
 	@$(OCAMLBUILD) $(OCAMLBUILDFLAGS) -clean
-	$(MAKE) clean-prefilter
 	@echo " Cleaned up working copy" # Note: ocamlbuild eats the first char!
 
 batteries.odocl: src/batteries.mllib src/batteriesThread.mllib
 	cat $^ > $@
 
-doc: prefilter batteries.odocl
+doc: batteries.odocl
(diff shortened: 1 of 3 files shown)


gasche:
The reason I am making a pull-request, instead of pushing directly, is that I think this move might be controversial as people are (understandbly) offput by the general lack of documentation of myocamlbuild.ml and the untidiness of Batteries' one. On the other hand #615 is arguably a minor issue as few people use `make coverage` on a regular basis (we would otherwise have discovered the failure earlier). This is your chance to protest if you think the cure is better than the disease.

rgrinberg:
+1 Maybe I'm underestimating the downsides but this honestly looks like a no brainer for merging.

UnixJunkie:
You fix something that's broken so it's a yes for me too.

hcarty:
Agreed. This looks like a good change.

PR #626, opened 2015-06-22
added BatSet.at_rank_exn

UnixJunkie:


gasche:
Here would be a suggested `.mli` documentation:

> [at_rank_exn i s] returns element at rank [i] in [s], that is the [i]-th element in increasing order (the [0]-th element being the smallest element).

UnixJunkie:
OK, done.

gasche:
Thanks, merged!

PR #628, opened 2015-07-23
expose `Format.asprintf` for V>=4.01, fixes #597

### LINKED ISSUE #597: Error: Unbound value Format.asprintf
It seems the Format.asprintf function which came in 4.01 is shadowed in Batteries.

If I have 

    Toploop.use_silently
             Format.err_formatter (Filename.concat (Findlib.package_directory
             "batteries") "battop.ml");;

in .ocamlinit, I get

    # #require "ppx_deriving.std";;
    # type d = float * float [@@deriving show];;
    Error: Unbound value Format.asprintf
    Hint: Did you mean sprintf or sprintf?

where I should be getting

    utop # type d = float * float [@@deriving show];;
    type d = float * float      
    val pp_d : Format.formatter -> float * float -> unit = <fun>
    val show_d : float * float -> bytes = <fun>

vogler:
Maybe it's better to backport it...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,5 +32,6 @@ src/batMarshal.mli
 src/batPrintexc.mli
 src/batPrintf.ml
 src/batPrintf.mli
+src/batFormat.mli
 src/batSys.mli
 src/batBigarray.mli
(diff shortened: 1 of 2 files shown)


gasche:
It is non-trivial to backport scanf/printf/format-related function to before 4.02 where they were rebased on GADTs, making them much more easy to implement. I'll merge this, thanks!

PR #632, opened 2015-08-17
added Map.update to reflect Set.update

UnixJunkie:
note that Map.add is already an update but not requiring the binding
to be already present

### DIFF
--- a/src/batMap.mli
+++ b/src/batMap.mli
@@ -89,6 +89,8 @@ sig
   val update: key -> key -> 'a -> 'a t -> 'a t
   (** [update k1 k2 v2 m] replace the previous binding of [k1] in [m] by
       [k2] associated to [v2].
+      This is equivalent to [add k2 v2 (remove k1) m], but more efficient
+      in the case where [k1] and [k2] have the same key ordering.
       @raise Not_found if [k1] is not bound in [m].
       @since NEXT_RELEASE *)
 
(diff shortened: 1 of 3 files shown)


gasche:
Two things are not great with this `Map.update`: it's not really a counterpart of `Set.update` (Set.update takes two elements of the set, and correspondingly `Map.update` could take two keys), and it is hardly different from `Map.modify`.

I think they still make sense in terms of performances compared to `Map.modify ... (fun _ -> v)`, and otherwise the code is of good quality (easy to review, it's exactly identical to the `Map.modify` code), so I'm tempted to include the PR nonetheless.

Are you planning to work on `at_rank`, `pop_min` and `pop_max` as well?

UnixJunkie:
I'll try to make Map.update more like the Set one then.
After this one, I'll try Map.at_rank.

UnixJunkie:
It should be OK now.
Plus: all new features from Set are now reflected in Map if the corresponding pull requests are accepted.

src/batMap.mli
```diff
@@ -86,6 +86,12 @@ sig
       [m], plus a binding of [x] to [y]. If [x] was already bound
       in [m], its previous binding disappears. *)
 
+  val update: key -> key -> 'a -> 'a t -> 'a t
+  (** [update k1 k2 v2 m] replace the previous binding of [k1] in [m] by
+      [k2] associated to [v2].
```
gasche: I would add the following sentence: "This is equivalent to [add k2 v2 (remove k1) m], but more efficient in the case where `k1` and `k2` have the same key ordering."

UnixJunkie:
ok, done

PR #633, opened 2015-08-18
Map.pop_min_binding

UnixJunkie:
reflection of BatSet.pop_min into Map

### DIFF
--- a/src/batMap.ml
+++ b/src/batMap.ml
@@ -105,6 +105,20 @@ module Concrete = struct
     | Node (l, _, _, _, _) -> min_binding l
     | Empty -> raise Not_found
 
+  let get_root = function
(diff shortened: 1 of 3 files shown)


UnixJunkie:
If this is OK, I will send the pop_max equivalent

UnixJunkie:
I sent pop_max also, finally.

gasche:
There is a conflict between this pull-request and #634 as both define the `get_root` function. Could you rebase this pull request on top of the current trunk, and check that it still compiles and passes the tests?

UnixJunkie:
I'll try.

UnixJunkie:
This is here: https://github.com/ocaml-batteries-team/batteries-included/pull/637

PR #634, opened 2015-08-25
Map.at_rank_exn

UnixJunkie:


gasche:
There was a compilation problem with this pull request which I fixed in b7ecb8f39cdcf499b94e595f3f34d62e9f5ac86d .

PR #635, opened 2015-08-25
corrected many warnings

UnixJunkie:


src/batPervasives.ml
```diff
@@ -163,7 +163,7 @@ let rec dump r =
       opaque "abstract"
     | x when x = Obj.custom_tag ->
       opaque "custom"
-    | x when x = Obj.final_tag ->
+    | x when x = Obj.custom_tag ->
```
gasche: This change is not correct, we need to keep `Obj.final_tag` here.

UnixJunkie:
I corrected it.

UnixJunkie:
this should have been merged

gasche:
I had mixed feeling about this change, sorry for not giving feedback earlier. My experience with string migrations is that they require somewhat delicate design choices of which functions to expose with type `string`, and which to expose with type `bytes`. This pull-request does not tackle any of these issues, because it is only designed to silence comments, not to actually make code usable with `safe-string` (where it would still fail to compile). I think it is a good first step (and thus I agree we can merge it now), but it is important to understand that it is only the simplest part of the iceberg. The goal is to have a `safe-string`-aware codebase.

gasche:
Also, one think that should be checked before merging this patch is the status of `Bytes` in Batteries today -- I don't have all the details in my head right now, and a check would be useful. Are the references to `Bytes` that I see in this patch to the BatBytes modules (which should do the right thing), or to the `Bytes` module of the standard library? In the latter case, does the code still compile with pre-Bytes version of OCaml such as 4.01.0 or 4.00.1? It should thanks to the `bytes` compatibility, but there is always a risk of things going wrong somehow.

TL;DR: I think the change can be merged if you have checked that it doesn't break compilation on one of the older OCaml versions.

UnixJunkie:
How far are we supposed to be compatible? 3.12.1 ?

gasche:
As far as what is advertised in README.md (the bound there should be consistent with a bound in the `opam` file, but currently there is no version constraint in OPAM; I'll try to fix it now).

Right now README.md says 3.12.1. We could decide to change if we feel that supporting too-old releases adds an undue maintenance burden. A good policy is to at least always support the OCaml version included in the latest Debian stable distribution (it was 3.12.1 for a very long time, but I think it just moved to a 4.x). I don't think there are any good reasons for dropping 3.12.1 for now -- we have 498 version-dependent lines of code, and only 10 of them apply to 3.12.1 specifically.

For this particular change, just checking *one* of the pre-Bytes version would be more than enough.

gasche:
This PR is superseded by the just-merged #651 , except for the BatLazyList work that would require use of conditional compilation.

PR #644, opened 2015-11-04
changed NEXT_RELEASE tags to the upcoming version number

UnixJunkie:


gasche:
Because we are adding a heap of new features (and not just fixing bugs), I think the next release version should be 2.4, rather than 2.3.2.

UnixJunkie:
can this one be merged too?

gasche:
Sorry, I missed the fact that the version number had been fixed. Thanks!

PR #646, opened 2015-11-06
synchronize BatText.nsplit to match BatString.nsplit (fixes #561)

### LINKED ISSUE #561: BatText.nsplit does not favour lack of spaces.
If there is no space in the string, nsplit does not work.

```
> nsplit (of_string "1,2,3,4") (of_string ",") |> List.map to_string;;
string list = ["1"]   
> nsplit (of_string "1, 2, 3, 4, ") (of_string ",") |> List.map to_string;;
string list = ["1"; " 2"; " 3"; " 4"]

### PR DESCRIPTION
**Pull request by gasche**
Supersedes #645 .

The BatText.nsplit implementation seems to suffer from a off-by-one
offset error. The present commit changes its implementation to reuse
exactly the more battle-tested implementation of BatString.split.

A consequence of this change are two small changes of behavior, that
corresponds to non-specified aspects of BatText.nsplit and arguably
fix bugs:
- nsplit now raises an (Invalid_argument ...) exception if the separator passed is empty (splitting on it makes no sense)
  (the previous implementation would loop endlessly)
- on two consecutive occurrences of the separate, nsplit now inserts an empty text in the sequence:
  `nsplit "a//b/c" "/" ` is `["a"; ""; "b"; "c"]` rather than `["a"; "b"; "c"]` as previously
  This ensures that for any `sep` and `str`, we have
    `join sep (split str sep)) = str`

### DIFF
--- a/src/batText.ml
+++ b/src/batText.ml
@@ -878,31 +878,50 @@ let rsplit (r:t) sep =
     avoid a call to [List.rev].  *)
 let nsplit str sep =
   if is_empty str then []
-  else let seplen = length sep in
-    let rec aux acc ofs = match
-      try Some(rfind_from str ofs sep)
-      with Not_found -> None
-    with
-    | Some idx ->
-      (* at this point, [idx] to [idx + seplen - 1] contains the
-            separator, which is useless to us on the other hand,
-            [idx + seplen] to [ofs] contains what's just after the
-            separator, which is what we want*)
-      let end_of_occurrence = idx + seplen in
-      if end_of_occurrence >= ofs then
-        aux acc (idx - 1) (*We may be at the end of the string*)
+  else if is_empty sep then invalid_arg "nsplit: empty sep not allowed"
+  else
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by gasche** [comment]
@UnixJunkie , I would be interested in your opinion in this approach to fixing #561 . Note that I had to change one of the existing unit-tests, which was

```
  nsplit (of_string "OCaml, the coolest FP language.") (of_char 'o') \
    |> List.map to_string = ["OCaml, the c"; "lest FP language."]
```

and now has an extra empty string

```
  nsplit (of_string "OCaml, the coolest FP language.") (of_char 'o') \
    |> List.map to_string = ["OCaml, the c"; ""; "lest FP language."]
```

UnixJunkie:
I am perfectly happy wit the new behavior and semantics. It looks indeed more correct.

gasche:
As you are the official "release manager" for the coming release, I will let you press the "Merge" button if you feel like it.

PR #647, opened 2015-11-06
Fixhashtbl

### LINKED ISSUE #609: Segmentation fault when using Hashtbl?
Dear all,

first of all, I would like to thank you for this great piece of software.

I suspect a bug in the functorial version of the Hashtbl module:
The following code results in a segmentation fault on my system
(64 bit Debian, OCaml 4.02.1 and Batteries Included 2.3.1):

```ocaml
open Batteries

module Item = struct
  type t = int

  let hash t = t
  let equal = (=)
end

module H = Hashtbl.Make (Item)

let _ =
  let h = H.create 7 in
  H.replace h (-4611686018427387904) [];
  H.find_default h (-4611686018427387903) []
```

It seems as if Hashtbl does not handle a hash value that equals min_int correctly 
on all instances (note the identity function for the hash operation in the Item module).

gasche:
batHashtbl: fix a hash computation bug introduced by 4.01 (fixes #609)
    
Martin Neuhäußer reports an important bug in the handling of
Hashtbl.Make hashing function in Batteries, caused by the
compatibility code to support both pre-4.01 and post-4.01 hash table
layouts. This bug is severe: it can turn into a segfault for specific
hash values (min_int in the reproduction example).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,6 +170,13 @@ qtest-byte-clean:
 qtest-byte: qtest-byte-clean
 	@_build/$(QTESTDIR)/all_tests.byte
 
+qtest-native-clean:
+	@${RM} $(QTESTDIR)/all_tests.ml
+	@${MAKE} _build/$(QTESTDIR)/all_tests.native
+
+qtest-native: prefilter qtest-native-clean
+	@_build/$(QTESTDIR)/all_tests.native
+
 qtest-clean:
 	@${RM} $(QTESTDIR)/all_tests.ml
 	@${MAKE} _build/$(QTESTDIR)/all_tests.$(EXT)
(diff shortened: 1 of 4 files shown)


UnixJunkie:
For traceability, I would have liked to see the exact code reported by the user (or at least recognize some of the lines and values he was using).
test_issue_609_1 indeed crash my toplevel but test_issue_609_2 doesn't and I don't know if it should.

gasche:
The literal values used by @martin-neuhaeusser are 64-bits `min_int` and `-max_int`. They cannot be used as-is, because they would raise a compile-time error on 32bits OCaml implementation. You are absolutely right that being more explicit in comments would be helpful.

- issue 1 is the segfault reported in the [first message](https://github.com/ocaml-batteries-team/batteries-included/issues/609#issue-66824302) of #609, that occurs with the current implementation of Batteries
- issue 2 is [the code](https://github.com/ocaml-batteries-team/batteries-included/issues/609#issuecomment-94851550) that was proposed as a counter-example to the correctness of my first patch draft, which was wrong. It does not fail with either the existing implementation, or the one proposed in this (corrected) patch, but I kept it as a regression test.

PR #650, opened 2015-11-10
Lazy.from_val replaced by new lazy_from_val to be compatible with 3.12.1

UnixJunkie:


gasche:
I think there is something wrong with this PR. On my machine, the current batteries master compiles fine with 3.12.1 (`opam config exec --switch=3.12.1 -- make all`). `man Lazy` says that `Lazy.from_val` is only available since 4.00, but `Lazy.lazy_from_val`, which is used in Batteries, seems to have always been available (and is now deprecated). In any case, I think the right replacement for (`Lazy.lazy_from_val v`) would be (`lazy v`), which is simpler than your re-implementation and is probably optimized reasonably by recent OCaml versions. the Lazy manual says something weird on that, but as neither of us are Lazy experts, let's do the simplest possible thing if we have to. But I don't think we do in this case.

UnixJunkie:
It also compiles fine for me now, I don't know what went wrong with my git repository.

PR #652, opened 2015-11-12
clean must also remove the qtest-generated file all_tests.ml

UnixJunkie:
or when you run 'make test' you can get errors which are outdated
for example after you do an opam switch to test with a different compiler


UnixJunkie:
How about this one?
I had the problem several times after an opam switch and even a make clean,
that there was still a left over file generated by qtest.
I here propose to remove that file upon make clean.

gasche:
I haven't had time to look at the details, but I'll trust you that it is fine. Thanks!

PR #657, opened 2015-12-16
Adds make functions in BatTuple

thizanne:
TupleN.make takes n arguments and builds a n-tuple.

No big stuff, but they can be useful e.g. as parameters of other functions, especially since `( , )` is no valid syntax.

Maybe `make` is not the better name: I also have in mind `build` and `create`, and have no idea which one (or which other one) is better.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,11 @@
 Changelog
 ---------
 
+## NEXT_RELEASE
+
+- BatTuple: add Tuple{N}.make : 'a1 -> ... -> 'an -> 'a1 * ... * 'an
+  (Thibault Suzanne)
+
 ## v2.4.0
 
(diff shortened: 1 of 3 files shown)


gasche:
The patch seems fine and `make` is a good name (in the stdlib `create` is for creating uninitialized structures). Could you add an entry to the changelog? This was suggested by @UnixJunkie to decrease the work at release time. Write what you would like users to read in the release announcement (short, but lists all the new functions).

UnixJunkie:
:+1:

thizanne:
What is supposed to be in the version line ? I assumed `NEXT_RELEASE` was fine.

gasche:
Yes, this is fine, thanks!

PR #661, opened 2016-02-11
Add cmp argument to some PSet functions

rixed:
Every PSet function creating a new set should allow to set the comparison
function.

### DIFF
--- a/opam
+++ b/opam
@@ -17,4 +17,5 @@ remove: [["ocamlfind" "remove" "batteries"]]
 
 depends: [
   "ocamlfind" {>= "1.5.3"}
+  "qtest" {test & >= "2.2"}
 ]


gasche:
I think that the reason why this was not done is that we hoped to have common module interfaces across various collections, in particular with `of_list`, `of_enum` functions etc. with the same types for `List`, `Array`, `Set` etc. This has proven not to work in the past (it's just too hard to obtain this kind of consistency), so this should not be an obstacle to this particular change.

That said, I'm a bit worried of the risk of breaking user code (adding optional arguments tend to break code that uses said functions without fully applying them). Any second opinion on this proposed change?

rixed:
Without the possibility to have another comparator than the generic one I think PSet.of_xyz are useless but as placeholder to ensure some functorial shenanigan can take place in some distant code base.
It seems to me it's more likely some code out there is bugged because some PSet are using the generic compare that shouldn't.

gasche:
Given that our next release is going to be major, it can break interfaces. I would be in favor of adding those arguments to the interface. Would you reopen and maybe propose a pull request?

PR #662, opened 2016-02-11
Add missing declaration of PSet.find

rixed:


gasche:
I think this omission suggests that nobody was using `PSet` through its explicit module (the `BatSet.PSet.foo` functions are also available as `BatSet.foo` which may have users), making it a good candidate for removal in Batteries v3.

rixed:
> the BatSet.PSet.foo functions are also available as BatSet.foo which may have users

Not sure what you mean here.
The BatSet functions are not exactly the same as the ones from PSet, since the ones from PSet work with a Set.PSet.t that's slightly more elaborate then the ones from BatSet.t. Do you suggest that nobody uses the additional feature of storing the comparison function within the set itself?
I, for one, needed this, for a set that was "half-polimorphic", the polimorphic part being ignored by the custom cmp fonction. I think this is a legitimate use case.

PR #669, opened 2016-03-04
Add min and max operations for all numeric types

### LINKED ISSUE #664: BatInt64 doesn't seem to contain a 'max' or 'min' function
Is there a reason for this?

atrieu:
Should fix #664.

Not sure how the tests are supposed to work, I kind of copy pasted some stuff around.
Didn't add tests when I just directly used the operation provided by the standard library, like `min_big_int` for example.


gasche:
I'm not fond of having `min,max` functions on the complex numbers: we do provide `compare` for convenience, but there is no reasonable total order on complex numbers and I'm not sure we should try to pretend there is one. I would rather remove those, and let users implement their own nonsensical operation if they want to.

The compiler-distributed standard library uses

```
let min x y = if x <= y then x else y
let max x y = if x >= y then x else y
```

it is important to use the exact same formulation, as it is necessary to guarantee consistency. For example, `min (0.) (-0.)` will return different results under your proposed implementation and the stdlib's.

atrieu:
I removed min and max for the complex numbers, but they will still be available through their ``Numeric`` interface.

And about the implementation for the floats, I just moved the one that was *already* present in the file into the ``BaseFloat`` module, so I guess the bug was already present but I fixed it now ? ;)

UnixJunkie:
@atrieu would be nice to move this forward. It's in conflict currently so cannot be merged.

UnixJunkie:
@atrieu would be nice to move this forward. It's in conflict currently so cannot be merged.

UnixJunkie:
Cannot be merged; nobody moved so closing.

Issue #519, opened 2014-01-21
List.sum fails on empty list


c-cube:
Strangely, `List.sum` fails on the empty list (instead of returning 0), but `Enum.sum` returns 0 in this case (same for `fsum`).

I guess this will have to wait until 3.0, but still, it's strange.

UnixJunkie:
Shouldn't this go in our upcoming release?

gasche:
I think we should keep it for the next major version push (3.0), because it breaks backward-compatibility. It is not a bugfix or a new feature, but something that improves consistency at the cost of changing a *specified* behavior.

You *could* decide to make the coming release a major release, but I suspect that it's a bit ambitious after our failure to get releases out in the past few month. I would be in favor of getting a release that is as polished and small-scope as possible now, to put a better release cycle on its rails, and later decide when we want our next major release and what to have in it -- for example some of @c-cube more ambitious inter-module dependency cuts.

UnixJunkie:
I just hit this bug, again...

rixed:
@UnixJunkie : there is no such function as `prod` or `fprod` it seems.

UnixJunkie:
@rixed yes, they might be useful. Be careful that such functions could also be reflected in BatArray.
sum/fsum also are in BatArray. For consistency, we might want to return 1 (or 1.0) in the empty data structure case and document it.

Issue #445, opened 2013-09-10
no BatFilename in batteries?


UnixJunkie:
Since we are an overlay of the stdlib, I was thinking
there would be one such module.
Even if it is just an alias for the one from the stdlib.

UnixJunkie:
Hendrik Boom asked for Filename.split_extension on caml-list. I think we should have it.

UnixJunkie:
And I should move my ass about this issue.

Issue #443, opened 2013-09-09
split for BatArray


UnixJunkie:
I'd like a split for arrays.
It would do roughly what split does for sets.
PRECONDITION: the array is sorted in increasing order

gasche:
`split` was just merged in master, will be part of 3.0.

rixed:
Reopening as we have too many Array.splits now: this one, that's been waiting on the v3 branch for a while and for a reason that eludes me, and another one added on master more recently that mimick List.split in that it splits an array of pairs into a pair of arrays.

To solve this conflict of name, I suggest @UnixJunkie  and @c-cube go fight a pistol-duel next Sunday morning and we keep the survivor's implementation, or I merely rename this split into something like "pivot_split", because that's what it evokes to me, unless someone has a better suggestion?

UnixJunkie:
Your renaming is fine.

UnixJunkie:
Fixed in v3.

PR #670, opened 2016-03-23
Promote to 4.03

kayceesrk:
The PR allows `batteries` to be built with `4.03.0+beta1`. The main change is to replace the use of custom `result` type with `Pervasives.result`.

### DIFF
--- a/src/batArray.mliv
+++ b/src/batArray.mliv
@@ -82,7 +82,9 @@ external make : int -> 'a -> 'a array = "caml_make_vect"
     If the value of [x] is a floating-point number, then the maximum
     size is only [Sys.max_array_length / 2].*)
 
-val make_float : int -> float array
+##V>=4.3##val make_float : int -> float array
+##V=4.2##external make_float : int -> float array = "caml_make_float_vect"
+##V<4.2##val make_float : int -> float array
 (** [Array.make_float n] returns a fresh float array of length [n],
     with uninitialized data.
 
--- a/src/batArray.mliv
+++ b/src/batArray.mliv
@@ -82,8 +82,7 @@ external make : int -> 'a -> 'a array = "caml_make_vect"
     If the value of [x] is a floating-point number, then the maximum
     size is only [Sys.max_array_length / 2].*)
 
-##V<4.2##val make_float : int -> float array
-##V>=4.2##external make_float : int -> float array = "caml_make_float_vect"
+val make_float : int -> float array
 (** [Array.make_float n] returns a fresh float array of length [n],
     with uninitialized data.
 
--- a/src/batGc.mli
+++ b/src/batGc.mli
@@ -34,18 +34,18 @@
 type stat = Gc.stat =
   { minor_words : float;
     (** Number of words allocated in the minor heap since
-        the program was started.  This number is accurate in
-        byte-code programs, but only an approximation in programs
-        compiled to native code. *)
+       the program was started.  This number is accurate in
+       byte-code programs, but only an approximation in programs
+       compiled to native code. *)
 
     promoted_words : float;
     (** Number of words allocated in the minor heap that
-        survived a minor collection and were moved to the major heap
-        since the program was started. *)
+       survived a minor collection and were moved to the major heap
+       since the program was started. *)
 
     major_words : float;
     (** Number of words allocated in the major heap, including
-        the promoted words, since the program was started. *)
+       the promoted words, since the program was started. *)
 
     minor_collections : int;
     (** Number of minor collections since the program was started. *)
@@ -62,7 +62,7 @@ type stat = Gc.stat =
 
     live_words : int;
     (** Number of words of live data in the major heap, including the header
-        words. *)
+       words. *)
 
     live_blocks : int;
     (** Number of live blocks in the major heap. *)
@@ -78,8 +78,8 @@ type stat = Gc.stat =
 
     fragments : int;
     (** Number of wasted words due to fragmentation.  These are
-        1-words free blocks placed between two live blocks.  They
-        are not available for allocation. *)
+       1-words free blocks placed between two live blocks.  They
+       are not available for allocation. *)
 
     compactions : int;
     (** Number of heap compactions since the program was started. *)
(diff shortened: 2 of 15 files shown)


gasche:
Thanks a lot. I'm worried by the fact that this change is a breaking change. Users would have to update their `Bad` constructor (Batteries-local result type) into `Error` (upstream result type), and there is no way to write code that is compatible with both versions.

I tried to see if there would be an upstream solution to not have this break code, see [MPR#7102](http://caml.inria.fr/mantis/view.php?id=7102&nbn=3), but it looks like this will not happen.

I think the best move is probably to accept that this is a breaking change, and take this as an occasion to bump Batteries' major version number. It does break user code, but it is a small breakage that is easy to fix by a global replacement. The other way would be to contortionate ourselves to maintain two separate `result` types (our own and upstream's), but that means more complex compatibility with other libraries using upstream `result`, etc.

src/batArray.mliv
```diff
@@ -82,8 +82,7 @@ external make : int -> 'a -> 'a array = "caml_make_vect"
     If the value of [x] is a floating-point number, then the maximum
     size is only [Sys.max_array_length / 2].*)
 
-##V<4.2##val make_float : int -> float array
-##V>=4.2##external make_float : int -> float array = "caml_make_float_vect"
+val make_float : int -> float array
```
gasche: Why did you feel the need for this change? I think that the idea was to follow the interface as present upstream so that the cast `BatArray : module type of Array` would succeed. (I'm not sure it does actually work, though, they may be order-of-fields issues anyway.)

src/batGc.mli
```diff
@@ -153,11 +159,17 @@ type control = Gc.control =
...
         can be better for programs with fragmentation problems.
         Default: 0. @since 3.11.0 *)
-  }
+
+    window_size : int;
+    (** The size of the window used by the major GC for smoothing
+        out variations in its workload. This is an integer between
+        1 and 50.
+        Default: 1. @since 4.03.0 *)
+}
```
gasche: We should probably export this field under a conditional test so that users can compile their code against the new Batteries but older OCaml version. Same thing for other `GC` primitives below. I can do the conditional stuff.

gasche:
I was a bit too dark on the backward-compatibility story: once users have converted their code from `BatPevasives.result` to `Pervasives.result`, they can use the backward-compatibility package for `result` that non-Batteries users will use as well. (I don't remember precisely whether it already exists and where it is.)

src/batArray.mliv
kayceesrk: Sorry. I was too casual there to delete the versioning. It appears the `make_float` interface has changed in 4.03 [0]. I've committed a fix to the interface that restores the original behavior. Not sure if it is a good idea to refer to the deprecated `Array.float`. Thoughts?

[0] https://github.com/ocaml/ocaml/blob/trunk/stdlib/array.mli#L55-L62

src/batGc.mli
kayceesrk: Agreed. Thanks!

kayceesrk:
Regarding the backwards compatibility, I think a major version bump that goes with the new compiler version is a reasonable idea. I am assuming there hasn't been further progress on the mantis issue.

c-cube:
Wouldn't a major version bump be the occasion to integrate other breaking changes? 0:-)

gasche:
@c-cube In theory, yes, in practice I'm sceptical that we will have time to integrate said breaking changes in a nice way in time for the 4.03 release. I'm very constrained right now, so I'd rather do something minimalistic.

What are the current proposals for breaking changes?

UnixJunkie:
I just remember of this one:
https://github.com/ocaml-batteries-team/batteries-included/issues/519

gasche:
And thanks to your helpful tagging work, we can find many other such issues:
  [for-next-major-release](https://github.com/ocaml-batteries-team/batteries-included/labels/for%20next%20major%20release)

I think we should integrate everything there, because they are small enough. #578 and #519 are clearly good ideas that mean little change for our users (but will definitely break backward-compat for the users of `Enum.combine`; I'm ok with this). I'm not sure why #443 was delayed to a next release but it seems fine, and #445 is more of a minor feature request that we may or may not implement depending on time and motivation.

c-cube:
for the record, I think batteries should rely on the `result` compatibility package, and guarantee that `BatResult.t = Result.result`. This is the correct choice in the long term (when everyone uses the standard `result`) and users will have to change their code anyway, whether the type alias is there or not.

gasche:
That's consistent with the fact that we use the `bytes` compatibility package. Any help in making this happen is warmly welcome.

I think that I'll merge the present PR soon, to get us started for the major release to happen not too far from the 4.03 release. Implementing the explicit use of the compatibility-package would be a warmly welcome change on top of it; if someone want to help, they can already start by working on top of @kayceesrk's promote-to-4.03 branch.

avsm:
Any ETA on this? This is blocking the compilation of OASIS on 4.03

gasche:
Merging in the master branch for pinners we can do quickly (I was planning to do this during the week, actually I was planning for 4.03 to happen later this week). Preparing the major release is a bit more work; hopefully this can done in this week or the next.

thelema:
Don't forget uchar module now in 4.03

c-cube:
I know it's too late, but would it make sense, for some next major release, to change `BatIO` so it becomes compatible with the standard types? I think we had arguments on that and that (almost?) no one was using `BatIO` because it doesn't play well with other libraries…

UnixJunkie:
@c-cube maybe for the next minor release

c-cube:
impossible, it would be breaking…

gasche:
I'm thinking of using the deprecation-warning facilities to decrease the use of IO and Enum over the 3.x series.

c-cube:
Fair enough ✔

UnixJunkie:
I also start to need batteries on 4.03.0. I have several packages depending on batteries that I'd like to test in 4.03. Tell me if I can help with some things.

gasche:
@UnixJunkie thanks for the offer!

There are two general things that would help:

- flux on IRC suggested trying to adapt the present PR to *not* replace `BatPervasives.result` with `result`. The idea is to have an intermediate step that is not good for the long-term (long-term we want to replace), but does not break existing Batteries' user code. We could release that intermediary step as a 2.x version, and have a 3.x at the same time with the replacement. Unfortunately, the present PR bundles several changes in the commits, and it's not trivial to separate the Result change from other changes we need to compile under 4.03 anyway (the Gc changes for example). So, to summarize: can one start from the present branch and have a PR that does *not* replace the result type, yet compile under 4.03? (One would need to change the way `Pervasives` is included in `BatPervasives` to avoid the name conflict.)

- There are all the PRs hanging around that we said we should consider for 3.0, with the [for-next-major-release](https://github.com/ocaml-batteries-team/batteries-included/pulls?q=is%3Apr+is%3Aopen+label%3A%22for+next+major+release%22) tag (thanks for helping trianging those!). We need to go over them, check if some changes were still pending and blocking merge (eg. add missing tests), and possibly rebase if they conflict with the current state of master. Rebasing is usually simple but a bit tedious, it would already help. (I think you have commit rights on the current repo, so you could push to a branch of the same name here if you do this).

On my side I need to finish the `-safe-string` work to have it in a state ready for 4.03. If nobody volunteers, I'll try to do the intermediary 2.x patch as well, but only next.

gasche:
Some progress report:
- in the branch [v2](https://github.com/ocaml-batteries-team/batteries-included/tree/v2) of the central repository there is a version that compiles under 4.03 but does not remove BatPervasives.result. I'll try to make a 2.x release off it soon.
- I have a port of `String.Cap` in the safe-string (still needs a bit more polish)

gasche:
I finished making the [v2](https://github.com/ocaml-batteries-team/batteries-included/tree/v2) branch pass the compatibility test, which verifies that all exported interfaces are strict supersets of those of the standard library -- this means extending the `.mli` files to support new 4.03 functions, and backporting them when it is easy to do so. The standard Queue module has a new internal representation in 4.03, so that also required some work.

I tested that our testsuite passes on all major OCaml versions from 3.12.0 to 4.03.0. I think that now only the work of actually putting a release out is left (hunting for NEXT_RELEASE mentions, etc., following [the release howto](https://github.com/ocaml-batteries-team/batteries-included/blob/master/howto/release.md)).

gasche:
The minor release is made, see ocaml/opam-repository#6515 .

What remains to be done is to integrate the final part of the present PR, namely the replacement of `BatPervasives.result` by `Pervasives.result`. This will be for the upcoming major release. As I write these lines, I notice that there are also some part of `gc.mli` that were in @kayceesrk's patch that I have forgotten to include in the work for 2.5.0, so these should also be included in the next release.

ghulette:
Is this still being worked on?

UnixJunkie:
@ghulette if you are worried, the current batteries compiles fine on recent OCaml versions.

ghulette:
Yes it works ok for me, I'm not filing an issue. Just wondering what the status is. From this thread it seems like at one point the idea was to cut a new major version with backward breaking changes, incorporating stuff like this. Is that still on the roadmap?

UnixJunkie:
Yes, but we are scarce on manpower, so things advance slowly.

c-cube:
Given [the current implementation](https://github.com/ocaml-batteries-team/batteries-included/blob/1e80fcf0c61e90d44bfb5af29e8f5d3f4c06e667/src/batInnerPervasives.mlv#L46) I think the issue still stands.

UnixJunkie:
I am closing this because it seems outdated (batteries works fine with 4.03).
This PR is too big.
Smaller PRs are more likely to be merged quickly.

PR #673, opened 2016-04-20
`-safe-string` compilation for Batteries

gasche:
In preparation for the upcoming 3.0 release, I have been working on switching Batteries to be usable with `-safe-string`. While the change is not compatibility-breaking for people that do not use `-safe-string` in their libraries yet (we are only replacing some instances of `string` by `Bytes.t` in the interfaces; and sometimes adding another function that works on the other type), it would break compatibility for people using Batteries and `-safe-string` together.

I would of course appreciate any review or comment, on the interface choices in particular.

The rest of this message discusses a detail point, `String.Cap`.

This module uses a phantom type discipline to distinguish read-only, write-only and read-write string, but the phantomness of it relies on defining all of these as `string`. I could try to turn this into a GADT that would hold either a `string` or a `bytes` depending on the intended usage, but this would change the interface slightly, as users would have to declare at string creation time what the usage mode is going to be (whereas this is inferred fully automatically). Also it would introduce some minimal amount of packing, against the current claim that using `Cap` has no performance impact whatsoever.

What I have done for now is to disable `String.Cap` module altogether. One could agree that it is less relevant in presence of the `string/bytes` distinction, and could be dropped from 3.0. I'm not convinced; I think that the users that were careful enough to use `Cap` in the past (we know that for example @jpdeplaix uses `Exceptionless`, which is similar in intended audience) would want to be able to write code polymorphic over the writeability of the string, and `string/bytes` does not allow this -- they would still feel a need for the more expressive interface.

An option we would have is to not have `String.Cap` in 3.0, and loudly advertise that we plan to add it back in future versions if users have a need for it and contact us about it. In particular, having a user at hand could help clarifying how the interface should evolve to meet their needs.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,16 @@
 Changelog
 ---------
 
+## v2.8.0 (minor release)
+
+This minor release supports the -safe-string mode for OCaml
+compilation, enforcing a type-level separation between (immutable)
+strings and mutable byte sequences.
+
+- support -safe-string compilation
+  #673
+  (Gabriel Scherer)
+
 ## v2.7.0 (minor release)
 
 This minor release is the first to support OCaml 4.05.0. As with
--- a/_tags
+++ b/_tags
@@ -10,3 +10,5 @@ true: package(bytes), warn_-3, bin_annot
 ".git": -traverse
 "examples": -traverse
 <src/batOpaqueInnerSys.*>: opaque
+true: safe_string
+
--- a/src/batBase64.ml
+++ b/src/batBase64.ml
@@ -67,7 +67,7 @@ let encode ?(tbl=chars) ch =
   in
   let output s p l =
     for i = p to p + l - 1 do
-      write (String.unsafe_get s i)
+      write (Bytes.unsafe_get s i)
     done;
     l
   in
--- a/src/batBig_int.mliv
+++ b/src/batBig_int.mliv
@@ -345,7 +345,7 @@ val nat_of_big_int : big_int -> Nat.nat
 val big_int_of_nat : Nat.nat -> big_int
 val base_power_big_int: int -> int -> big_int -> big_int
 val sys_big_int_of_string: string -> int -> int -> big_int
-val round_futur_last_digit : string -> int -> int -> bool
+val round_futur_last_digit : Bytes.t -> int -> int -> bool
 val approx_big_int: int -> big_int -> string
 ##V>=4.3##val round_big_int_to_float: big_int -> bool -> float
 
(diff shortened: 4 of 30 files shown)


c-cube:
I didn't look at the code, but couldn't `String.Cap` use `bytes` where it used to use `string`?

UnixJunkie:
Not completely on the point, but to me it is fine to break compatibility
between major releases. Of course, if you break everything every-time there
is a major release, users will not stay.

gasche:
@c-cube Yes, that's also an option. I think it's better than trying to dive into GADTs head-first.

@UnixJunkie indeed, given that we decided to break compatibility in the next release, it is fine to break compatibility. We should still understand and document well what we break and why. But on the question of `-safe-string` in particular I'm not worried about backward-compatibility (except maybe for the Cap thing). Today it is morally wrong to rely on the types of Batteries when using `-safe-string` anyway, as Batteries may mutate `string` parameters; so users of `-safe-string` will be glad that we break compatibility, as it makes their code safer. (And disabling `-safe-string` gives an easy way to remain backward-compatible with their pre-change codebases.)

jpdeplaix:
@gasche for the record, I don't use Batteries anymore. I replaced it with @c-cube 's Containers.
But yes, apart for that, I have used the `Exceptionless` module but not the `Cap` module tho.

gasche:
@c-cube I started working on that but it's still not-so-convincing and a lot of work for two reasons:

1. The only correct types you can give for going in and out of the `cap` interface are `_ t -> bytes`, `bytes -> _ t`, and `string -> [< `Read] t`. Because of the `read_only : _ t -> [`Read] t` function, there is no safe way to return a `string` from a `Cap.t`, which means that the `to_string` function will always have a counter-intuitive type -- really, it should be deprecated.

2. Most of the `Cap` implementations are not implemented for Bytes. I could hack my way through wrapping each operation in conversion function, but the "right" thing to do is to define a `BatInnerByteString` module that define operations for both datatypes and shares as much code as possible (plus, if all the code is internal to a single module it is safer to reason about ownership and use unsafe conversions), but that is a fair amount of work.

I'm hesitating between just having `String.Cap.{of,to}_string` just manipulate `Bytes.t`, and use conversion functions all over the place (the easiest option, still some work, kind of ugly for end-users), and just removing the module altogether as I initially proposed.

jpdeplaix:
Any news on this ?

gasche:
We I think there would be nothing wrong with rebasing the current PR and merging it (thanks for the ping!), but a bit of work remains to fix the `-safe-string` build.

gasche:
I just rebased this PR on top of the current master. I hope to merge in the following week(s), and have a release before 4.06.0+beta and the release, to allow users to test their own packages.

gasche:
Merged! Apologies for the delay. We have to prepare a release now to be 4.06-ready.

PR #675, opened 2016-04-21
Use Base.compare for Infix operators

314eter:


gasche:
There was an unfortunate duplication of work with #676, apologies for not reacting earlier.

@rixed's patch is slightly better as it comes with tests, and also converts the dangerous use of `if y >= x` at the beginning of the function. I'll merge/rebase the patches and close the issue.

Thanks a lot for your contribution -- report and fix.

gasche:
I merged the fix and added a Changelog entry. Thanks!

PR #676, opened 2016-04-23
Fix #674.

### LINKED ISSUE #674: Infix operators (--) and (---) don't work for Big_int
The bug is introduced in commit 3129fbbac0a7888d9ff53b5673ee61d99547e200. The module `MakeInfix`

```ocaml
module MakeInfix (Base : NUMERIC_BASE) :
  Infix with type bat__infix_t = Base.t = struct

  type bat__infix_t = Base.t
  let ( + ), ( - ), ( * ), ( / ), ( ** ) = Base.add, Base.sub, Base.mul, Base.div, Base.pow
  let ( -- )  x y = BatEnum.seq x Base.succ ( (>=) y )
  let ( --- ) x y =
    if y >= x then x -- y
    else BatEnum.seq x Base.pred ((<=) y)
end
```

uses comparisons `(>=)` and `(<=)` which are not yet defined by `MakeCompare`.

rixed:
Range infix operators must use specific comparison operators rather than
generic ones that may not work for some types (such as Big_ints).

### DIFF
--- a/src/batBig_int.ml
+++ b/src/batBig_int.ml
@@ -157,3 +157,18 @@ let print out t = BatIO.nwrite out (to_string t)
     BatIO.to_string print (power_int_positive_int 10 31) = "10000000000000000000000000000000"
     BatIO.to_string print (power_int_positive_int (-10) 31) = "-10000000000000000000000000000000"
   *)
+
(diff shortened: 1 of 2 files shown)


gasche:
Merged, thanks!

PR #677, opened 2016-04-23
s/QTest2Lib/qcheck/

rixed:
Not sure why but qcheck library have been renamed some time ago. As a consequence I could not
make test. This patch change findlib requirement name to pull the proper library. I'm curious why this
was not done already ; maybe batteries dev are supposed to use an older version of qcheck? Or
maybe there is a better way to configure this?

### DIFF
--- a/qtest/_tags
+++ b/qtest/_tags
@@ -1,3 +1,3 @@
 true: threads, debug
-<all_tests.*>: package(oUnit), package(QTest2Lib)
+<all_tests.*>: package(oUnit), package(qcheck)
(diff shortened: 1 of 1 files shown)


c-cube:
Oh, apologies. I'm responsible for the change of name and forgot to push the same patch in batteries.

gasche:
Thanks! I noticed that the qtests were broken on my machine after merging @c-cube's patch, but I wasn't quite sure whether it was a local problem or a global issue.

PR #679, opened 2016-04-23
Add cmp argument to some PSet functions

rixed:
Every PSet function creating a new set should allow to set the comparison
function.


gasche:
Before I merge, could you maybe add a mention in the ChangeLog?

gasche:
Merged in master, will be in 3.0. Thanks!

PR #682, opened 2016-06-08
Documentation: fix attachement issues in batBigarray

Octachron:
This very minor pull request fixes some attachment issues in the documentation comment in `batBigarray.mliv`: some documentation comment were erroneously attached to the last constructor rather than the whole type. This PR fixes this problem by adding an empty comment `(**)` after the last constructor.


gasche:
Excellent, thanks!

PR #684, opened 2016-07-02
Add sort functions to BatTuple modules

a12n:
Hello,

This adds functions to sort {2,3,4,5}-tuples. The functions are implemented with sorting networks.

Is it a good idea to have such functions in the lib?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,21 @@ Changelog
 
 ## v2.5.0
 
+Batteries 2.5.0 is a minor release whose main change is that it is
+compatible with the newly released OCaml 4.03.
+
+Note that Batteries 2.5.0 is compatible with older OCaml releases as
+well, and provides back-ported versions of most standard library
+functions made available in 4.03 only. For example,
+BatString.uppercase_ascii is usable under all OCaml versions.
+
+If the documentation of a Batteries function says
+  @since 2.5.0
+then it is available under all supported OCaml version (3.12.1 and up).
+If it says
+  @since 2.5.0 and OCaml 4.03.0
+then it is only available under OCaml 4.03.0.
+
 - BatTuple: add Tuple{N}.make : 'a1 -> ... -> 'an -> 'a1 * ... * 'an
   (Thibault Suzanne)
 - BatBig_int: fix sequence operators (--), (---) to avoid polymorphic comparison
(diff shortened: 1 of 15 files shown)


gasche:
What is a good use-case for those sorting functions?

(I usually tend to think of tuples as heterogeneous structures, with the various members of different types. `sort` requires homogeneous tuple so it is arguably a special case, where tuples are rather seen as statically-known-length lists. Nothing fundamentally wrong with that special-casing but I wonder when it is used in practice.)

Do we really want to use comparison networks here? Moving data around is very cheap, and the operation that may be costly in general is the comparison operation. For `Tuple3` for example, your implementation always perform three comparisons, while two comparisons suffice in many cases. As for comparison networks, there is no known uniform algorithm to compute optimal solutions, but the so-called Ford-Johnson method is known to be optimal for sizes below N=13 so it could be used/specialized here.

(Thanks a lot for the randomized test-cases, I wouldn't trust the code otherwise.)

a12n:
Probably, there aren't that many real-world applications.

I've used these sort functions mostly in educational contexts (e.g., returning ordered tuple of array indices from {2,3,4}SUM problems, returning ordered pair of distances along a ray intersecting two planes). `Tuple5.sort` was added mainly for consistency, I've never had need for such a thing.

You're certainly right about `Tuple3`. I've thought that sorting networks would be more understandable in code than a hierarchy of `if`s. Implemented with `if`s, reusing `sort` for smaller tuples, it looks cleaner. Also, it may use less comparisons than a sorting network (although, seems like it will use more comparisons than a corresponding sorting network in the worst case). So, it might be better way to implement these functions.

(Thanks a lot for the `qtest` itself, it's like life changing, I've only used `ounit` and `qcheck` directly before.)

gasche:
I don't think that decision trees do more comparison than sorting networks in the whole case; they have more flexibility (they can do perform different operations depending on the result of a test, instead of having a statically-determined continuation), so they can be strictly better. In the `Tuple3` case it is clear for example.

a12n:
I've changed `Tuple3.sort` implementation. But 24-leaf decision tree for `Tuple4` already seems too tedious to write.

UnixJunkie:
This should be broken into smaller, independent PRs.
Also, before sending a new PR, discuss about the functionality in advance using an issue
in the issue tracker.

PR #685, opened 2016-07-22
Remove dead code and fix documentation.

johanneskloos:
The documentation for BatInt.of_string was misleading:
The current implementation throws Failure, not Invalid_argument.

This is because the implementation being referred to was dead code,
which is being removed in this commit.

### DIFF
--- a/src/batInt.ml
+++ b/src/batInt.ml
@@ -95,9 +95,6 @@ module BaseInt = struct
   external to_int : int -> int = "%identity"
 
 
-  let of_string x =
-    try int_of_string x
-    with Failure err -> raise (Invalid_argument err)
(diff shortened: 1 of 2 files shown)


johanneskloos:
This patch makes BatInt compatible with BatFloat, BatInt32 and BatInt64. If you prefer, I also have a patch that works the other way round, changing the exception behviour of {Batint,BatFloat,BatIn32,BatInt64}.of_string to raise Invalid_argument instead of failure.

gasche:
Using `Failure` is the correct choice to remain consistent with the compiler distribution's standard library: `int_of_string`, or `Int32.of_string`, both use `Failure` rather than `Invalid_argument`. Thanks!

PR #686, opened 2016-08-03
Weighted choice

gasche:
`BatRandom.weighted_choice : ('a * float) Enum.t -> 'a`

### DIFF
--- a/src/batRandom.ml
+++ b/src/batRandom.ml
@@ -73,6 +73,51 @@ let enum_char () = BatEnum.from char
 
 let choice e = BatEnum.drop (int (BatEnum.count e)) e; BatEnum.get_exn e
 
+let weighted_choice e =
+  let list, total_weight =
+    BatEnum.fold (fun (acc, total) ((_, w) as elem) ->
+        if w < 0. then
+          Printf.kprintf invalid_arg
+            "weighted_choice does not accept negative weights (here %f)"
+            w;
+        elem :: acc, total +. w) ([], 0.) e in
+
+  assert (total_weight >= 0.);
+  if list = [] then
+    invalid_arg "weighted_choice: cannot choose from empty enumeration";
+
+  let pick = float total_weight in
+  let rec find level =
+    (* [find level rest] invariant:
+         pick > level
+         level + (sum of weights of rest) = total_weight *)
+    function
+    | [] ->
+      (* this case can never arise, as it would imply
+         that (level + 0 = total_weight) and thus (pick > total_weight),
+         but (pick < total_weight) by definition of pick. *)
+      assert false
+    | (elem, w) :: rest ->
+      let next_level = level +. w in
+      if pick <= next_level then elem
+      else find next_level rest
+  in find 0. list
+(*$T weighted_choice
+  try ignore (weighted_choice (BatList.enum [1, 3.; 1, -2.])); false \
+    with Invalid_argument _ -> true
+  try ignore (weighted_choice (BatList.enum [])); false \
+    with Invalid_argument _ -> true
+  weighted_choice (BatList.enum [true, 1.]) = true
+  weighted_choice (BatList.enum [true, 1.; false, 0.]) = true
+  let choice () = weighted_choice (BatList.enum \
+    [0, 1.; 1, 10.; (-1), 0.; 2, 1.]) in \
+  Array.init 100_000 (fun _ -> choice ()) \
+  |> BatSet.of_array \
+  |> BatSet.elements \
+  |> (=) [0; 1; 2]
+*)
+
+
 (* Reservoir sampling algorithm (see for instance
    http://en.wikipedia.org/wiki/Reservoir_sampling)
 
--- a/src/batRandom.mli
+++ b/src/batRandom.mli
@@ -115,6 +115,25 @@ val choice : 'a BatEnum.t -> 'a
     This function only works on finite enumerations with
     less than 2{^30} elements.*)
 
+val weighted_choice : ('a * float) BatEnum.t -> 'a
+(** [weighted_choice e] takes pairs of elements and their weight, and
+    returns a randomly-chosen element according to its weight: the
+    probability of each element to be chosen is equal to its weight
+    divided by the sum of all weights. The underlying algorithm is
+    sometimes called the "fortune wheel".
+
+    For example, [weighted_choice (List.enum [(true, 2.); (false, 4.)])]
+    will return [false] twice more often than [true], as it has double weight.
+
+    This function only works on finite enumerations with
+    less than 2{^30} elements.
+
+    @raise Invalid_argument if one of the weights is negative
+    @raise Invalid_argument if passed an empty enumeration
+
+    @since NEXT_RELEASE
+*)
+
 val multi_choice : int -> 'a BatEnum.t -> 'a BatEnum.t
 (** [multi_choice n e] returns an enumeration of [n]
     randomly-chosen elements of [e]. *)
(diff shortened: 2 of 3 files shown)


UnixJunkie:
This line could be more explicit:

 +  let pick = float total_weight in 

I guess pick is a random number drawn from a uniform distribution.
But I'm not sure, even from reading stdlib's Random module doc.

gasche:
Yes, `Random.float x` returns a uniform float between 0 (included) and x (excluded),  which is what is needed here.

In my actual use-case for this function, I am actually trying to draw `n` elements according to their weights. I wonder if that is not the primitive we should expose to users -- it's inefficient to implement it on top of this version, as the user needs to filter out the element explicitly, and then calling the function again recomputes the total weight. It could have type `('a, float) enum -> 'a enum`, enumerating as many distinct elements as possible.

UnixJunkie:
Maybe it should be in two steps: 1st you prepare the weighted list. 2nd you draw from it as many times as you wish.

cakeplus:
Strangely, this PR includes several unrelated commits. Did someone rebase published history on `master`? Not nice.

src/batRandom.mli
```diff
@@ -115,6 +115,24 @@ val choice : 'a BatEnum.t -> 'a
...
+    its weight divided by the sum of all weights.
+
+    For example, [weighted_choice (List.enum [(true, 2.); (false, 4.)])]
+    will return [false] twice more often than [true], as it has double weight.
+
+    This function only works on finite enumerations with
+    less than 2{^30} elements
+
+    @raise Invalid_argument if one of the weights is negative
+    @raise Not_found if passed an empty enumeration
```
cakeplus: Perhaps `Invalid_argument` in this case is more reasonable? `Not_found` is typically used for signaling _search failure_, and I see no search here.

src/batRandom.ml
```diff
@@ -73,6 +73,37 @@ let enum_char () = BatEnum.from char
...
+          Printf.kprintf invalid_arg
+            "weighted_choice does not accept negative weights (here %f)"
+            w;
+        elem :: acc, total +. w) ([], 0.) e in
+
+  if list = [] || total_weight <= 0. then raise Not_found;
+
+  let pick = float total_weight in
+  let rec find level = function
+    | [] -> assert false
```
cakeplus: I'd prefer an informative message:
```
invalid_arg "weighted_choice does not accept an empty enumeration"
```

src/batRandom.ml
```diff
@@ -73,6 +73,37 @@ let enum_char () = BatEnum.from char
...
+let weighted_choice e =
+  let list, total_weight =
+    BatEnum.fold (fun (acc, total) ((_, w) as elem) ->
+        if w < 0. then
+          Printf.kprintf invalid_arg
+            "weighted_choice does not accept negative weights (here %f)"
+            w;
+        elem :: acc, total +. w) ([], 0.) e in
+
+  if list = [] || total_weight <= 0. then raise Not_found;
```
cakeplus: Same here:
```
invalid_arg "weighted_choice requires at least one positive weight"
```

UnixJunkie:
I think this algorithm is called a fortune wheel usually.
Might be nice to see it mentioned somewhere.

src/batRandom.ml
gasche: This case is in fact statically known to be impossible, hence the `assert false`. I added a comment to justify this impossibility. It is always good practice to do so, thanks for the remark.

gasche:
I rebased the PR and took the review comments into account. Thanks! If another round passes, I think I'll merge the PR.

(I'm not fond of using `Enum.t` as the input type of these functions, but using something else for `weighted_choice` (`array` would make the most sense, I think, at least for `choice`) would create inconsistencies with `choice`, so it's an interface change best left for a future major version I think.

src/batRandom.ml
cakeplus: I just meant that an informative message here would be better than checking for `list = []` earlier. As for assertion being unreachable, it was quite obvious, as the algorithm is trivial. To be honest, I found the new explanation comment much more difficult to understand than the code, though additional documentation will do no harm, I guess.

src/batRandom.ml
```diff
@@ -73,6 +73,51 @@ let enum_char () = BatEnum.from char
...
+    with Invalid_argument _ -> true
+  weighted_choice (BatList.enum [true, 1.]) = true
+  weighted_choice (BatList.enum [true, 1.; false, 0.]) = true
+  let choice () = weighted_choice (BatList.enum \
+    [0, 1.; 1, 10.; (-1), 0.; 2, 1.]) in \
+  Array.init 100_000 (fun _ -> choice ()) \
+  |> BatSet.of_array \
+  |> BatSet.elements \
+  |> (=) [0; 1; 2]
+*)
```
cakeplus: This will fail once in a while. There should be an award (a congratulatory message) for someone who triggers this : - )

cakeplus:
The new version is all good. I'm for merging.

src/batRandom.ml
cakeplus: Hmm, the upper bound for the probability of this jackpot is (11/12)^100000 : - )

src/batRandom.ml
```diff
@@ -73,6 +73,51 @@ let enum_char () = BatEnum.from char
...
 let choice e = BatEnum.drop (int (BatEnum.count e)) e; BatEnum.get_exn e
 
+let weighted_choice e =
+  let list, total_weight =
+    BatEnum.fold (fun (acc, total) ((_, w) as elem) ->
+        if w < 0. then
+          Printf.kprintf invalid_arg
+            "weighted_choice does not accept negative weights (here %f)"
+            w;
+        elem :: acc, total +. w) ([], 0.) e in
```
murmour: This is what I'd do in 2017, because folds destroy readability:
```ocaml
let total = 0.0 in
e |> List.iter (fun ((_, w) as elem) ->
  if w < 0. then
    Printf.kprintf invalid_arg
      "weighted_choice does not accept negative weights (here %f)"
      w;
  total := !total +. w);
```

It also seems that reversing [e] into [list] is unnecessary.

src/batRandom.ml
```diff
@@ -73,6 +73,51 @@ let enum_char () = BatEnum.from char
...
+
+  assert (total_weight >= 0.);
+  if list = [] then
+    invalid_arg "weighted_choice: cannot choose from empty enumeration";
+
+  let pick = float total_weight in
```
murmour: `total_weight` is already a `float`.

UnixJunkie:
@gasche there are still some (very small) unresolved comments.
This is a cool contribution, I even have use cases for this.

rixed:
Going to address those unresolved comments and update the PR unless objections, in which case consider closing this maybe?

UnixJunkie:
After some (long) time, my opinion has evolved.
I sometimes need such a function.
I would never use an Enum for that.
And while compiling the list of weighted things would be probably done once,
drawing from it would be done many times.
Which probably means that a non trivial data structure is needed to accelerate this,
like an interval tree or something.
Also, I don't think making BatEnum and BatIO more featureful is a very good idea, because one day those two modules are just going to fly away.

UnixJunkie:
My opinion is that such a function should be high performance.
Hence, not for "BatEnum the uggly".
If you disagree, please reopen.

murmour:
I think the idea was to make this function data-structure-agnostic, with Enum acting as a generator used to fill other structures (lists, arrays, trees, etc). But I agree that extending Enum is undesirable. It's legacy technology that should be phased out in favor of Seq, and its functionality is better be frozen.

PR #694, opened 2016-12-05
Make split_at lazy.

01mf02:
gasche's contribution from #688.

PR #695, opened 2016-12-08
Fix compilation of LazyList

thizanne:
A recent change (#694) made `LazyList.split_at lazy`, by using `drop`. The result does not compile on a fresh clone on my machine, because `drop` is defined after.

This fix attempt simply moves `drop` before `split_at` in the module implementation. Nothing else is changed.


gasche:
Indeed, sorry for missing that in the original PR.

PR #696, opened 2016-12-08
Add List.{map2i,iter2i}

thizanne:
I could use a `List.map2i` function, so here it is. For the sake of consistency, I also added `iter2i`, but I don't have enough time×motivation tonight to add a `_2i` version of each `_2` function of `List` (partly because it feels like a process which I'm not sure whether it terminates).

I propose to add `map2i` and `iter2i` anyway since they are written, and to wait until someone needs them (or until my shame takes over my laziness) to write the other ones.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,6 +15,7 @@ Changelog
   (Michael Färber, Gabriel Scherer, Thibault Suzanne)
 
 - Add `List.{map2i,iter2i}`
+  #696
   (Thibault Suzanne)
 
 ## v2.5.3
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,9 @@ Changelog
   #694
   (Michael Färber, Gabriel Scherer, Thibault Suzanne)
 
+- Add `List.{map2i,iter2i}`
+  (Thibault Suzanne)
+
 ## v2.5.3
 
 Batteries 2.5.3 synchronizes library functions with OCaml 4.04+beta2,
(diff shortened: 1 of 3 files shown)


gasche:
I think these functions are fine, but could you add inline tests for them? You should test cases with two empty lists, two one-element lists, two several-element lists, and at least one size-mismatch failure mode.

thizanne:
Is this ok ? I cannot run `make test` on my machine (#698). Btw, I rebased.

thizanne:
Now that #698 is fixed on my machine, I ran the tests, they pass.

UnixJunkie:
Good to see some young blood helping with batteries. :)

gasche:
One thing that is problematic with the `List.foo2` functions is that they document that they raise `Different_list_size`, but in fact raise `Invalid_argument`. The documentation of existing functions is already wrong -- there is, in fact, no `Different_list_size` exception in `BatList`, there is one only in `BatLazyList`. I'm not sure that it is a good idea to introduce a new function with wrong documentation, but I'm not sure that it would be a good idea to use a different exception either. Maybe you could weaken the specification of those functions by saying `raise Different_list_size or Invalid_argument`. Long-term, we have to think about whether we want to change the exception or the specification (now or for the next major version, v3).

UnixJunkie:
The stdlib's List.map2 throws Invalid_argument when list sizes differ.
I guess batteries should do the same since we are an overlay of the stdlib.

thizanne:
Done. I only fixed the documentation for iter2i and map2i -- as you said, let's think about the other ones in another PR.

gasche:
I would be happy to merge, but could you please include the PR number in the changelog entry?

thizanne:
Done !

PR #703, opened 2017-01-06
RFD: extensions for BatMap.Make (do not pull!)

jrochel:
Hello everyone!

First of all I want to express my appreciation for the library. I use it everyday and I'm very grateful to everyone involved!

I've been using `BatMap.Make` quite a lot and I found some functions are sorely missing and implemented them in a module `BatMapPlus.Make`. I'd like to hear your comments on whether you consider these functions also useful, which one you would like to see incorporated and which not.

If a concensus is reached I'd go on to create a proper patch for the `BatMap.Make` module.

### DIFF
--- a/src/batMapPlus.ml
+++ b/src/batMapPlus.ml
@@ -1,0 +1,54 @@
+open BatPervasives
+module Enum = BatEnum
+module List = BatList
+module Opt = BatOption
+
+module type ORD = BatInterfaces.OrderedType
+
+(** adds some missing functions to BatMap.Make *)
+module Make (Key : ORD) = struct
+
+  include BatMap.Make (Key)
+
+  let diff : 'a t -> 'b t -> 'a t = fun m1 m2 ->
+    let by = fun _ x y -> match x, y with Some x, None -> Some x | _ -> None
+    in merge by m1 m2
+
+  let union : 'a. ?zero:'a -> ('a -> 'a -> 'a) -> 'a t -> 'a t -> 'a t =
+    fun ?zero f -> merge @@ match zero with
+      | Some zero -> let d = Opt.default zero in fun _ l r -> Some (f (d l) (d r))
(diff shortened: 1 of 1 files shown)


jrochel:
I'd really like to see functions like `union` and `intersect` in `BatMap.Make`. Doesn't anyone agree?

thizanne:
Some `intersection` function that takes a function `f` to merge the values makes sense and goes in the direction of providing ready-to-use interfaces. We probably could also have `intersection_{left,right}` that simply choose some argument's binding to build the new table. I'm also approve a `diff` function, but I would intuitively expect it to be a symetric difference (we could also have `diff_left` and `diff_right`).

It's a bit more ambiguous with `union` : I think the right way to do it, if it's to be done, is to have a merging function `f : 'a -> 'a -> 'a` that is only called if a key is present on both maps (that should be properly documented, of course). Again, `union_left` and `union_right` are obvious. I'm not sure about functions using `zero` parameters : I feel like we don't really gain ease of use from simply using `merge`.

Concerning `of_list`, one purpose of `enum` is precisely to avoid having n² conversion functions between data structures. One could argue that lists are common enough, but I think we should rather encourage people to use enums where appropriate (and replace them with sequences, when someone has time to work on v3). 

We also already have something like `dump`, it's the `print` function. It can seem a bit more tedious at first (doesn't return a `string`), but it's more flexible and in fact probably not longer to use when you get to actually use the string you generated.

gasche:
Be careful: the standard library has a `union` function in `Map.S` now (since 4.03.0), so if we reuse the same name it is important to also use the same interface, to try to preserve the property that we are a conservative extension of it.

jrochel:
Thank you for your comments! I expect to get around to work on this in a few weeks. As recommended by UnixJunkie in #719 I plan on making a series of small pull requests starting with the least controversial functions.

Until then!
Jan

PR #705, opened 2017-01-10
Added BatResult.{map,map_both}

### LINKED ISSUE #704: BatResult does not have a map function
I know that BatResult exports a `map_default` function, but I was looking for something similar to the `map` in BatOption. Is there any particular reason why something similar to the flowing is missing from BatResult.ml?

    let map f = function
      | Bad e -> Bad e
      | Ok v  -> Ok (f v)

If there isn't any particular reason, could we have this added?

ifazk:
This resolves #704

### DIFF
--- a/src/batResult.mli
+++ b/src/batResult.mli
@@ -38,6 +38,16 @@ val get : ('a, exn) t -> 'a
 *)
 val default: 'a -> ('a, _) t -> 'a
 
+(** [map f (Ok x)] returns [Ok (f x)] and [map f (Bad e)] returns [Bad e].
+    @since 2.5.4
+*)
+val map : ('a -> 'b) -> ('a, 'c) t -> ('b, 'c) t
+
+(** [map_both f g (Ok x)] returns [Ok (f x)] and [map_both f g (Bad e)] returns [Bad (g e)].
+    @since 2.5.4
+*)
+val map_both : ('a -> 'b) -> ('c -> 'd) -> ('a, 'c) t -> ('b, 'd) t
+
 (** [map_default d f r] evaluates to [d] if [r] is [Bad] else [f x]
     when [r] is [Ok x]
     @since 2.0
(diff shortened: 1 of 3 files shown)


src/batResult.mli
```diff
@@ -38,6 +38,16 @@ val get : ('a, exn) t -> 'a
...
 
+(** [map f (Ok x)] returns [Ok (f x)] and [map f (Bad e)] returns [Bad e].
+    @since 2.5.4
+*)
+val map : ('a -> 'b) -> ('a, 'c) t -> ('b, 'c) t
+
+(** [map_both f g (Ok x)] returns [Ok (f x)] and [map_both f g (Bad e)] returns [Bad (g e)].
+    @since 2.5.4
+*)
+val map_both : ('a -> 'b) -> ('c -> 'd) -> ('a, 'c) t -> ('b, 'd) t
```
gasche: I think this signature would be more readable with `'a1 -> 'a2`, `'b1 -> 'b2`.

src/batResult.mli
```diff
@@ -38,6 +38,16 @@ val get : ('a, exn) t -> 'a
 *)
 val default: 'a -> ('a, _) t -> 'a
 
+(** [map f (Ok x)] returns [Ok (f x)] and [map f (Bad e)] returns [Bad e].
+    @since 2.5.4
+*)
+val map : ('a -> 'b) -> ('a, 'c) t -> ('b, 'c) t
+
+(** [map_both f g (Ok x)] returns [Ok (f x)] and [map_both f g (Bad e)] returns [Bad (g e)].
+    @since 2.5.4
```
gasche: could you use `NEXT_RELEASE` instead of `2.5.4`, so that we can replace with the correct version (which may not be 2.5.4) at release time?

ifazk:
Fixed. Should I also change the ChangeLog file?

gasche:
Yes, please. Thanks!

ifazk:
Done!

gasche:
Thanks, I just merged.

PR #711, opened 2017-01-26
Common sets

UnixJunkie:
Here is a proposal for BatInt.Set.
If this is OK with most maintainers, I can duplicate the job
to provide the same for floats and strings.
As always, I am not sure about what should go into the .mli since
I use them so few.
Please check it (especially the two type lines).
Regards,
F.

### DIFF
--- a/src/batInt.ml
+++ b/src/batInt.ml
@@ -289,6 +289,10 @@ module Safe_int = struct
 
 end
 
+module Set = struct
+  include (Set.Make(BaseInt))
+end
+
 (*$T &
   Result.(catch (Safe_int.add max_int) max_int |> is_exn Number.Overflow)
   Result.(catch (Safe_int.add min_int) min_int |> is_exn Number.Overflow)
--- a/src/batInt.mli
+++ b/src/batInt.mli
@@ -387,4 +387,38 @@ module Safe_int : sig
   (** Equality function for integers, useful for {!HashedType}. *)
 
   val ord : t -> t -> BatOrd.order
+
 end
+
+module Set :
+sig
+  type elt = int
+  type t
+  val empty : t
+  val is_empty : t -> bool
+  val mem : elt -> t -> bool
+  val add : elt -> t -> t
+  val singleton : elt -> t
+  val remove : elt -> t -> t
+  val union : t -> t -> t
+  val inter : t -> t -> t
+  val diff : t -> t -> t
+  val compare : t -> t -> elt
+  val equal : t -> t -> bool
+  val subset : t -> t -> bool
+  val iter : (elt -> unit) -> t -> unit
+  val fold : (elt -> 'a -> 'a) -> t -> 'a -> 'a
+  val for_all : (elt -> bool) -> t -> bool
+  val exists : (elt -> bool) -> t -> bool
+  val filter : (elt -> bool) -> t -> t
+  val partition : (elt -> bool) -> t -> t * t
+  val cardinal : t -> elt
+  val elements : t -> elt list
+  val min_elt : t -> elt
+  val max_elt : t -> elt
+  val choose : t -> elt
+  val split : elt -> t -> t * bool * t
+  val find : elt -> t -> elt
+  val of_list : elt list -> t
+end
+(** Ready to use intset. *)
--- a/src/batFloatSet.ml
+++ b/src/batFloatSet.ml
@@ -1,1 +1,0 @@
-include (Set.Make(BatFloat))
--- a/src/batFloatSet.ml
+++ b/src/batFloatSet.ml
@@ -1,0 +1,1 @@
+include (Set.Make(BatFloat))
--- a/src/batIntSet.ml
+++ b/src/batIntSet.ml
@@ -1,0 +1,1 @@
+include (Set.Make(BatInt))
--- a/src/batIntSet.ml
+++ b/src/batIntSet.ml
@@ -1,1 +1,0 @@
-include (Set.Make(BatInt))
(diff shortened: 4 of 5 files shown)


src/batInt.ml
```diff
@@ -289,6 +289,10 @@ module Safe_int = struct
 
 end
 
+module Set = struct
+  include (Set.Make(BaseInt))
```
cakeplus: The outer parenthesis are unnecessary.

src/batInt.mli
```diff
@@ -387,4 +387,38 @@ module Safe_int : sig
   (** Equality function for integers, useful for {!HashedType}. *)
 
   val ord : t -> t -> BatOrd.order
+
+end
+
+module Set :
```
cakeplus: Why did you copy-paste this? Just make a reference:
``` ocaml
module Set: BatSet.S with type elt = int
```

src/batInt.mli
```diff
@@ -387,4 +387,38 @@ module Safe_int : sig
...
+  val cardinal : t -> elt
+  val elements : t -> elt list
+  val min_elt : t -> elt
+  val max_elt : t -> elt
+  val choose : t -> elt
+  val split : elt -> t -> t * bool * t
+  val find : elt -> t -> elt
+  val of_list : elt list -> t
 end
+(** Ready to use intset. *)
```
cakeplus: This comment is redundant. How can a set be not "ready to use"? :-)

src/batInt.ml
cakeplus: Also, I'd just write:
``` ocaml
module Set = Set.Make(BaseInt)
```

src/batInt.mli
thizanne: `Set.Make` is not ready to use, you need to instantiate it. I won't start a fight about one comment line, but that's probably what @UnixJunkie expressed.

cakeplus: Yeah, but this isn't `Set.Make` -- it's `Set.S`, which already has an appropriate docstring (it will be shown in the output of both ocamldoc and ocp-index).

Perhaps a better comment would be "A set of integers"?

UnixJunkie:
I prefer no comment rather than completely paraphrasing the code.

cakeplus:
On the proposal as a whole: I agree that it'd be useful to have pre-instantiated sets and maps for the common basic types. It would be great to have `Int.Set`, `Int.Map`, `String.Set`, `String.Map`, `Char.Set`, `Char.Map`.

UnixJunkie:
let's get the set issue settled first. Common maps could be the next story.

hcarty:
IIRC these modules used to exist a long time ago but were removed to limit the amount of cross-module dependencies in Batteries.  I'm not sure if that's still a concern or not but if it is these could live as separate top-level modules (`Int_set`, etc.), possibly using module aliases to make `Int.Set` available as an alias.

UnixJunkie:
It is still a concern.
Batteries might be split into several independent modules some day.
I am also in favor of separate top-level modules.
I am not in favor of aliases/duplication of things.
So @hcarty, you prefer Int_set to IntSet?
But there is already BatBitSet for example.
I was thinking BatIntSet would be a natural candidate; other maintainers have made other proposals.

cakeplus:
One argument in favor of new toplevel modules is that it'd be more convenient to provide alternative specialized implementations for them. For example, my personal implementation of `IntSet` is based on a Patricia tree, and for many (most? all?) of the use-cases it is much more performant than the default AVL-tree-based `Set`. For `CharSet` we could use a bitset. For `BatString` -- a Radix?

cakeplus:
> IIRC these modules used to exist a long time ago but were removed to limit the amount of cross-module dependencies in Batteries. I'm not sure if that's still a concern or not but if it is these could live as separate top-level modules (Int_set, etc.), possibly using module aliases to make Int.Set available as an alias.

Making it `Set.Int` instead of `Int.Set` would help with cross-module dependencies.

But why is this much of a concern? It's the compiler's (linker's) job to strip dead code from the binaries.

thizanne:
I was in favour of "int sets" being in the `Set` module as a "common instantiation of `Set.Make`", but the argument of specialised implementations makes sense indeed. But then I think it should be in `Int.Set` as the module hierarchy seems better organised this way  : modules or functions that are integers' business are expected to be found in `BatInt`. Plus, it's still the right way to pronounce it in english, cc @UnixJunkie ;) But that leaves us with the cross-dependency issue to solve.

thizanne:
Actually the cross-dependency exists because of `Set.to_array`, that relies on `DynArray`, that relies on `Int`. That seems very specific and I think we could change `to_array` without losing on the performance to leverage this issue.

UnixJunkie:
@thizanne if you can send a new version of Set.to_array, that would be cool.

thizanne:
Update: #724 fixes the cross-dependency and allows to have `Int.{Set,Map}`, and we can have the same for `Float`, but there is another cross-dependency for `Char` and `String` (that goes through `Printf` and `Buffer`). This one seems more problematic, and I still don't like the "let's add plenty of top-level modules" solution. I think my preferred solution is to have `Set.{Int,Char, ...}`, as explicit "common instantiations", and if we want to add, say, patricia-trees-based int sets at some point, add a new toplevel module then (that says something like "specialised set structure for integers - faster than Set.Int").

Unless someone objects I'm going to implement this.

UnixJunkie:
I don't object.
About the (better) specialized implementations; I just think it should replace the default
one if it is shown to be better in terms of performance for most (or at least the most useful) operations.

UnixJunkie:
Most of this is done by ffac0420a5e4b83239c3213547ed711acf0e2621.
@cakeplus if you want to provide some specialized implementation for some of the already
instantiated sets we now provide, please open a new issue.

PR #718, opened 2017-02-07
added Bat{List|Array}.shuffle

UnixJunkie:


gasche:
You should try to install the resulting version of `Batteries` and load it from a toplevel to run one of the new functions. I suspect that you need to add the new file to the `.mllib` (and maybe some other place?) for this to work.

(You can update pull-requests by just doing `git push --force` towards you personal github clone of the repository, no strong need to close a PR to open a new one.)

UnixJunkie:
I am testing it; I think the experience in the toplevel is not what people want.

gasche:
(The failure model I had in mind was not toplevel-specific: if I remember correctly from my own BatConcreteQueue mistakes, compiling a program using Batteries fails at link-time if the `mllib` is not complete.)

UnixJunkie:
I changed the semantic and ocamldoc strings.
Please have a look before accepting (eventually).

gasche:
I am not convinced by the change: why change away from optional arguments again, is there a good reason for this?

I think the change you want to make is just turning

```
?(s = Random.get_state ()) ... ->
```

into

```
?s ... ->
let s = match s with
| Some s -> s
| None -> Random.get_state ()
in
```

UnixJunkie:
In case no random state was provided, several successive calls to shuffle should move the global
random state forward.
In case a random state was provided, it is moved forward but not given back to the caller, which might be a problem.
I don't know how to detect with an optional argument that the argument was not passed; hence I reverted back to an option type.

UnixJunkie:
What I describe is also the behavior I observed in Core.List.permute.

UnixJunkie:
other proposal there: https://github.com/ocaml-batteries-team/batteries-included/pull/720

UnixJunkie:
done by 3c3bc7d8ea860c695d18d1934f4925cf523f8feb

PR #720, opened 2017-02-10
another proposal for shuffling lists and arrays

UnixJunkie:
I hope this one will go through at last ...

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,10 @@ Changelog
 
 ## NEXT_RELEASE (minor release)
 
+- added {BatList|BatArray}.shuffle
+  #702, #707
+  (Francois Berenger, Gabriel Scherer)
+
 - Clarification and improvements to the documentation
   #682, #685, #693
   (Florian Angeletti, Johannes Kloos, Michael Färber)
--- a/src/batArray.mliv
+++ b/src/batArray.mliv
@@ -533,6 +533,18 @@ val ord : 'a BatOrd.ord -> 'a array BatOrd.ord
     lexicographically for arrays of the same size.  This is a
     different ordering than [compare], but is often faster. *)
 
+val shuffle : ?state:Random.State.t -> 'a array -> 'a array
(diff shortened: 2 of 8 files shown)


gasche:
I realize now that I misunderstood the behavior of `get_state ()`: it does an independent copy of the generator state, instead of returning the same shared state as I assumed. This is weird, and arguably a design bug in the API.

With this in mind, I understand why the previous versions did not work as you expected. The newly proposed one does not have a nice behavior either, because it initializes new random state, and libraries should not do this (they should use the default state or the user-provided state). There are two solutions, and they both use the trick that when you take a function parameter as `?x`, *without* providing an option value, and you have `?x:foo` in the API, you get a variable x` of type `foo option` back.

Solution 1: exactly as you did in 9cc908a6f61aed97bd0b0b71c6ed0b20ce63ad52, select either `Random.State.int s` or `Random.int` depending on whether a state was passed or not.

Solution 2: exactly as I suggested before, but after shuffling is finished you look back at the state again: if no state was explicitly passed (so you know you have a copy of the global state), you call `Random.set_state` to propagate the changes.

Solution 1 is clearly the nicer of the two. What would be even better would be some `get_shallow_state` function that returns an alias of the global random state, instead of a copy, to use instead of `get_state`.

If that sounds difficult to you, I am willing to implement the patch myself and send a PR with the whole thing (that is, mostly your commits).

UnixJunkie:
I will not send more code concerning this feature request.
I am done with it.

UnixJunkie:
@gascher: "The newly proposed one does not have a nice behavior either, because it initializes new random state, and libraries should not do this".
In the current case; this is perfectly fine since this new random state is only used internaly to the function.

UnixJunkie:
please maintainers, make this move on.
Then we can probably make a new release.

gasche:
@UnixJunkie: I made the change I had in mind in  20fab777cf11d9ef19bbf04e19a283052a2bc846 , and submitted a new PR #721.

PR #721, opened 2017-02-10
Shuffle

gasche:
Subsumes #720, and reuses most of its code.

I need to add more tests to make sure that the behavior is as expected before we can merge. Any comment is, of course, welcome.

### DIFF
--- a/src/batArray.mliv
+++ b/src/batArray.mliv
@@ -533,6 +533,18 @@ val ord : 'a BatOrd.ord -> 'a array BatOrd.ord
     lexicographically for arrays of the same size.  This is a
     different ordering than [compare], but is often faster. *)
 
+val shuffle : ?state:Random.State.t -> 'a array -> 'a array
+(** [shuffle ~state:rs a] randomly shuffles in place the elements of [a].
+    The optional random state [rs] allows to control the random
+    numbers being used during shuffling (for reproducibility).
+
+    Shuffling is implemented using the Fisher-Yates
+    algorithm and works in O(n), where n is the number
+    of elements of [a].
+
+    @since NEXT_RELEASE
+*)
+
 val equal : 'a BatOrd.eq -> 'a array BatOrd.eq
 (** Hoist a equality test for elements to arrays.  Arrays are only
     equal if their lengths are the same and corresponding elements
--- a/src/batArray.mliv
+++ b/src/batArray.mliv
@@ -533,7 +533,7 @@ val ord : 'a BatOrd.ord -> 'a array BatOrd.ord
     lexicographically for arrays of the same size.  This is a
     different ordering than [compare], but is often faster. *)
 
-val shuffle : ?state:Random.State.t -> 'a array -> 'a array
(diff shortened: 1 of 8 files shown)


src/batInnerShuffle.ml
```diff
@@ -0,0 +1,12 @@
+let array_shuffle ?state a =
+  let random_int state n = match state with
+    | None -> Random.int n
+    | Some s -> Random.State.int s n in
+  for n = Array.length a - 1 downto 1 do
+    let k = random_int state (n + 1) in
+    if k <> n then
```
c-cube: you should put '(' ')' around this `then`, as the indentation suggests. Here the second `unsafe_set` will always run.

src/batList.mlv
```diff
@@ -1304,6 +1304,15 @@ let subset cmp l l' = for_all (fun x -> mem_cmp cmp x l') l
...
 *)
 
+let shuffle ?state l =
+  let a = BatInnerShuffle.array_shuffle ?state (Array.of_list l) in
+  Array.to_list a
+(*$T shuffle
+  let s = Random.State.make [|11|] in \
+  shuffle ~state:s [1;2;3;4;5;6;7;8;9] = [7; 2; 9; 5; 3; 6; 4; 1; 8]
+  shuffle [] = []
+*)
```
c-cube: since I think the remark above is actually a bug in the old implementation, maybe a random test that checks whether the new array is a permutation of the old one would be helpful.
Something like:

```ocaml
(*$Q
  Q.(array_of_size Gen.(2--15) small_int) (fun a -> \
    let a' = Array.copy a in \
    shuffle a'; \
    (Array.to_list a' |> List.sort Pervasives.compare) = \
     (Array.to_list a |> List.sort Pervasives.compare))
*)
```

`(*$Q
  Q.(small_list small_int) (fun l ->\

src/batInnerShuffle.ml
gasche: I don't think it actually will always run: it would have a hard time given that it uses a variable that is local to the branch ;-) I think the code is behaving as it should, but you are correct that multi-line conditionals should be bracketed to avoid confusion.

src/batArray.mliv
```diff
@@ -533,6 +533,18 @@ val ord : 'a BatOrd.ord -> 'a array BatOrd.ord
     lexicographically for arrays of the same size.  This is a
     different ordering than [compare], but is often faster. *)
 
+val shuffle : ?state:Random.State.t -> 'a array -> 'a array
```
gasche: Should the shuffle on array really have type `'a array -> 'a array` while it is done in-place (mutate the input and returns it)? I would rather expect `'a array -> unit` to better reflect the behavior. An external opinion would help.

UnixJunkie:
yeah, return a unit for Array.shuffle is better

gasche:
I added @c-cube's test that the result is a permutation of the input, and another test that all permutations may occur (on input arrays of size 5).

UnixJunkie:
thanks to all the contributors, especially for the property tests

PR #726, opened 2017-02-15
Fix a whitespace oddity in Changelog

thizanne:
This seems to be a "tabulation instead of spaces" thing. My editor (emacs with automatic whitespace-cleanup on save) always changes these two particular lines when adding a Change. I would personnally find it comfortable not to manually remove them from the commit every time (and keeping the automatic cleanup for my own editions), but I would also understand if we don't want to (half-)arbitrarily make this change.


cakeplus:
> My editor (emacs with automatic whitespace-cleanup on save) always changes these two particular lines when adding a Change.

I think it's a problem of your text editor. What if the setting was different? What if your editor replaced tabs with spaces like my Emacs does? When I have similar problems with on-save cleanup hooks, I simply add exceptions for particular files (or even groups of them).

Issue #709, opened 2017-01-23
we should have ready to use IntSet, StringSet, etc.


UnixJunkie:
Hello,
This was recently asked by someone else.
Since we are 'batteries included', I think we should provide
already made sets.
This was a good idea.

UnixJunkie:
could avoid the stupid code like this


    module IntSet = struct
      include Set.Make(Int)
      (* extend type with more operations *)
      let of_list = L.fold_left (fun acc x -> add x acc) empty
      let to_list = elements
    end

UnixJunkie:
It could avoid stupid code like this in one's project:

    module IntSet = struct
      include Set.Make(Int)
      (* extend type with more operations *)
      let of_list = L.fold_left (fun acc x -> add x acc) empty
      let to_list = elements
    end

UnixJunkie:
Apparently there is something to do when addind a new module in batteries.
Is there some myocamlbuild.ml or _tags file I should update?
I am trying to add src/batIntSet.ml for example.

thizanne:
Shouldn't `IntSet` be in `batSet.ml` ? `Set` would then provide polymorphic sets (`Set` module), a functor (`Set.Make`), and some common applications of the functor (`Set.IntSet` and so on).

UnixJunkie:
I would prefer to write IntSet directly: Set.IntSet is redundant.
Syntax is important and conciseness too to have beautiful code.

thizanne:
Then what about something like `Set.Int` (or `Int.Set`, again ?) ? Conciseness is cool but I don't think `IntSet` should be seen as a totally separated module from the `Int` and `Set` ones.

c-cube:
I think `Set.Int` (and `Map.String`) makes the most sense.

UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/pull/711

UnixJunkie:
done by ffac0420a5e4b83239c3213547ed711acf0e2621

Issue #692, opened 2016-11-17
Documentation: IntMap


01mf02:
In the documentation for batSet:
 
> You can also use predefined maps such as IntMap for maps with integer keys.

However, IntMap is nowhere defined. Still, a predefined IntMap would be very convenient to have!

gasche:
Correct, `IntMap` was removed in f26ecc8b9dcdef87e11236186194702c904167b9 in 2012.

On the other hand, it's fairly easy for users to redefine as `BatMap.Make(BatInt)` (or `Map.Make(Int)` if you `open Batteries`), and because it's an applicative functor two definitions in this form will return the same module (the types will be compatible). My personal inclination would be to replace this mention by an example of defining `IntMap` and `StringMap` that way. (One reason why is that it avoids having a dependency from the `Map` module to `Int` and `String`, which would be an issue if we decided one day to use balanced map in a `String` algorithm for example.)

Would you say that having `IntMap` predefined would really be better than `module IntMap = BatMap.Make(BatInt)`

01mf02:
I did not know about `Map.Make(Int)`, which is really handy.
I added your examples to #693 and think that this makes a definition of IntMap obsolete.

thizanne:
See #727

PR #727, opened 2017-02-15
Common Map and Set functor applications

thizanne:
This is yet another proposal for #709 and #712. It does add a dependency from `BatSet` and `BatMake` to the argument modules, but I think it's acceptable, as `Int`, `Float` and other number modules can probably be seen as foundation modules anyway (that won't depend on `Set`), and there already is a dependency from `Set` and `Map` to `Char` and `String` (so we don't actually add one here -- which could be seen as a problem).

I'm all ears about other common instantiations to add.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,6 +22,11 @@ Changelog
   #705
   (Ifaz Kabir)
 
+- Add {BatSet,BatMap}.{Int,Int32,Int64,Nativeint,Float,Char,String} as
+  common instantions of the respective `Make` functor.
+  #709, #712
+  (Thibault Suzanne, François Bérenger)
+
 - BatString: add `chop : ?l:int -> ?r:int -> string -> string`
   #714, #716
   (Gabriel Scherer, request by François Bérenger)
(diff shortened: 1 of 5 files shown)


UnixJunkie:
make test

    File "src/batMap.ml", line 1107, characters 20-31:
    Error: This expression has type
             ('a -> 'a -> int) -> 'a BatMap.Int.t -> 'a BatMap.Int.t -> int
           but an expression was expected of type
             ('a -> 'a -> int) -> ('a -> 'a -> int) -> int
           Type 'a BatMap.Int.t is not compatible with type 'a -> 'a -> int

UnixJunkie:
on the other hand, make install works and the new ready to use sets are visible and usable in the toplevel.
Once make test is repaired, I will vote in favour of including this very useful and user-friendly contribution.

thizanne:
Good catch. Let me know when it's ok to merge, I shall squash this properly.

thizanne:
Like, what about #692

UnixJunkie:
In my opinion, this is OK to merge. Tests are OK now.

UnixJunkie:
Not merged yet?
We should close this https://github.com/ocaml-batteries-team/batteries-included/issues/712 and this https://github.com/ocaml-batteries-team/batteries-included/issues/709 after.

gasche:
(I have a deadline today and another next Monday, so I will not personally participate in a merge decision for this. Feel free to merge yourself if you think this is the right thing to do.)

PR #728, opened 2017-02-16
Hashtbl of/to list

UnixJunkie:


thizanne:
Shouldn't `to_list` be named `bindings`, for consistency with `Map` ? Or at least be an alias. 

More broadly, I'm not sure we should provide these functions everywhere. The point of having enums (or sequences) is partly to have O(n) conversion functions between structures instead of O(n²). One could argue that lists are pervasives enough in OCaml that we should also convert to and from them, but I'm not sure we should encourage to use lists everywhere in a world where we have sequences.

I'm not against merging but this is probably a design that should be discussed at some point (maybe only for v3 though).

UnixJunkie:
Many data structures have a (to_list, of_list) pair of functions.
I'm OK if you want to add an alias.
We still have enums (unfortunately) and we don't have sequences yet (unfortunately).

Issue #658, opened 2015-12-17
get rid of the .mlv/.mliv files


UnixJunkie:

gasche:
For the record I made an experiment this week-end with a [cppo branch](https://github.com/gasche/batteries-included/tree/cppo) in my private repository. I converted two modules to use cppo instead of the local build/prefilter, [Array](https://github.com/gasche/batteries-included/commit/7ab5a65d56d275a558ecb47cd1a07e9da0a0caae) and [Bigarray](https://github.com/gasche/batteries-included/commit/588c54f8da7c2aa0f6429c3e0bd07eebbba11a12).

The results are mixed. The good things about cppo:
- it is standard; people know about it, they may have adapted their editor-mode configuration to recognize `.ml.cppo` extensions, etc.
- it has other features than filtering on the version that we might use in the future (but we don't need them for now): defines, inclusion, etc.
- it supports a richer language of version condition; in particular, "and" and "or" are definable, so you could enable some code on some like `(v > 4.0 && v < 4.2) ||| (v = 4.3.1)`, which we cannot do without duplication (or more development) with build/prefilter -- but we don't need it for now.

The bad thing is:
- it makes an additional dependency (but I'm fine with that)
- the syntactic overhead is actually larger (see the examples above), so the code could become *harder* to read
- it cannot prefilter line inside comments, which mean that comments with some part that are conditional on some versions would need to be duplicated.

All in all, I think "cppo is more standard" is the killer argument and it's *worth* converting, but it's also a lot of (boring) work and, as you can see, the benefits are not so clear-cut. I'm putting this on hold and concentrating on getting 4.03-ready releases.

If someone is interested in working on helping on that, please feel free to say something here.

UnixJunkie:
Are you forced to have X.ml.cppo ?
Can't you have X.ml and then tell the build system that this file needs preprocessing?

gasche:
Using a specific extension seems standard among `cppo` users (see [utop](https://github.com/diml/utop/tree/master/src/lib)) for example, but now that I think of it Jérémie seems to use `.cppo.ml`, which is better editor-wise and such, so I should probably use that. I'm fine with whatever the ocamlbuild plugin provided with `cppo` allows, but I personally like the idea of having an explicit mark somewhere that the code is not *quite* standard OCaml, but has some extra stuff. (We could use `.v.ml` for our stuff as well, I guess?)

UnixJunkie:
[...]

PR #729, opened 2017-02-17
Remove some deprecation warnings in Float for doc generation

thizanne:
I simply followed warning recommendations, I didn't add any annotation (maybe we should, at least some float functions in the compiler stdlib have more annotations than what we have). 

I don't really know the details of C interface, so it would be nice if someone can confirm that `"noalloc"` can be replaced with `[@@noalloc]` everywhere, and same for `"float"` and `[@@unboxed] [@@noalloc]`.

### DIFF
--- a/src/batFloat.mliv
+++ b/src/batFloat.mliv
@@ -81,8 +81,12 @@ val sub : float -> float -> float
 val mul : float -> float -> float
 val div : float -> float -> float
 
-external modulo : float -> float -> float = "caml_fmod_float" "fmod" "float"
-external pow : float -> float -> float = "caml_power_float" "pow" "float"
+external modulo : float -> float -> float = "caml_fmod_float" "fmod"
+##V<4.3##  "float"
+##V>=4.3##  [@@unboxed] [@@noalloc]
+external pow : float -> float -> float = "caml_power_float" "pow"
+##V<4.3##  "float"
+##V>=4.3##  [@@unboxed] [@@noalloc]
 
 val min_num : float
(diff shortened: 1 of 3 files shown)


gasche:
How does this change affect the OCaml versions that we support?

thizanne:
Didn't think about that. I'll keep this open and wait until we switch to cppo to do this properly.

UnixJunkie:
Yes, we need cppo and a macro to do this.

UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/658

UnixJunkie:
What is the ocaml version where "noalloc" became [@@noalloc] ?

gasche:
4.03.0

UnixJunkie:
@thizanne can you update your PR using the current preprocessor batteries uses?
It can already handle such cases.

thizanne:
Rebased and added some preprocessing. We still have a lot of warnings, but I don't plan to fix the other in this PR.

thizanne:
I didn't know / thought about that. I'll do the change. This will also easily fix some more warnings.

thizanne:
I pushed the suggested change, also fixing the same kind of warnings in Strings.

For the record the remaining warnings are the deprecated Pervasives and the deprecated mutability of `Gc` record fields ones. I'm still not sure what's the proper way of fixing them and don't plan to think about it right now.

UnixJunkie:
Thanks a lot for having updated the PR.
It should be merged very soon.

PR #732, opened 2017-02-20
added scripts/install_test.sh

UnixJunkie:


gasche:
`test_install` would be a better filename, otherwise it seems fine. Could you add a Makefile target that runs it (`make test-install`) and include it in the release howto?

UnixJunkie:
I did all that.
But then I saw in howto/release.md that someone added comments about make test-build-from-install.
Since make test-build-from-install is not fully automatic, I should we get rid of it.

UnixJunkie:
do we remove make test-build-from-install and include my proposal?

gasche:
Unfortunately, I think that "less automatic" is a bit better in this case, because it means that people know enough about how it works to orient themselves when it fails. With your script, if the test build breaks, people are in a different place and it can be confusing. Similarly, it installs for them, which means that it may trash their existing batteries install (in their common opam switch) if they are not careful about when they run it. I'm sorry I had forgotten that this was already done, but I think I prefer the existing way...

UnixJunkie:
I'm in favor of integrating my script.
Because I did it; it works and I know what it does.
We can leave the release doc as is but I am stil interested in having my automatic script to do the job for me.

howto/release.md
```diff
@@ -13,6 +13,9 @@ Make a release
   software in your main development switch, feel free to move to
   a fresh new switch to test this.)
 
+- instead of the previous, you can also run the fully automatic
+  'make test_install'
```
gasche: could you explicitly mention that it rebuilds, installs, and run a test in a temporary directory?

gasche:
Ok, then I'm fine with merging both. I mentioned a minor improvement to the release documentation (being more explicit about the inner workings decrease the chances of an user being surprised in a bad way), then you or I could merge the PR.

UnixJunkie:
superseded by https://github.com/ocaml-batteries-team/batteries-included/pull/743

PR #737, opened 2017-03-06
added BatList.favg

UnixJunkie:


gasche:
Could you write it with a `while` loop and a float reference?
This version allocates a floating box on each list element.

UnixJunkie:
done for favg

UnixJunkie:
done for fsum

gasche:
You may want to run a benchmark to compare your version with one with a `while` loop, such as:

```
let favg l =
  match l with
  | [] -> invalid_arg "Empty List"
  | x::xs ->
    let acc = ref x in
    let len = ref 1 in
    let rem = ref xs in
    let go = ref true in
    while !go do
        match !rem with
        | [] -> go := false;
        | x::xs ->
          acc := !acc +. x;
          incr len;
          rem := xs
      done;
      !acc /. float_of_int !len
```

UnixJunkie:
I have other things to do than running benchmarks.
If there are volunteers, please go ahead.
@gasche if you know that this is more efficient, then you can push such code. I personally don't write code that looks like the output of a compiler.

gasche:
Also, it would be nice if the failure case carried the function name in the exception message.

UnixJunkie:
I added function names to exception messages.

UnixJunkie:
this would be cool to have this in the next release

UnixJunkie:
I will send a link to some benchmarks.
Apparently: fsum_opt is 15% faster than fsum_naive.
favg_opt is 45% faster than favg_naive.

UnixJunkie:
cf. https://github.com/UnixJunkie/batlist_bench_01

UnixJunkie:
superseded by https://github.com/ocaml-batteries-team/batteries-included/pull/746

PR #741, opened 2017-03-16
BatteriesConcreteQueue.cmi and cmx were installed capitalized while o…

UnixJunkie:
…thers are consistently installed uncapitalized


gasche:
Nice catch!

PR #745, opened 2017-04-06
added BatList.frange

UnixJunkie:
should be quite useful to scientific programmers.
It's the second time that I identify that I need this in a project, so here it is.

### DIFF
--- a/src/batList.mli
+++ b/src/batList.mli
@@ -147,6 +147,19 @@ val range : int -> [< `To | `Downto ] -> int -> int list
     @raise Invalid_argument in ([range i `Downto j]) if (i < j).
     @since 2.2.0 *)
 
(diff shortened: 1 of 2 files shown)


gasche:
What's the reason for asking `To/Downto` instead of inferring it from whether `start <= stop` or `start >= stop`?

UnixJunkie:
I tried to match the interface of BatList.range (which I added long time ago).
Also, in scientific code it doesn't hurt to be very explicit about what you are trying to do.

gasche:
Oh, sure. Merged, thanks!

gasche:
I added a Changes entry in 4720636ae66a1cbdf9bb75201fef0bb755e3df05.

UnixJunkie:
thanks for the review and the changelog update

PR #746, opened 2017-04-07
added BatList.favg and faster BatList.fsum

UnixJunkie:


UnixJunkie:
Once this is integrated, I can look at the release next week.

gasche:
The code and tests look good, thanks.

I think that it is wrong for `sum` and `fsum` to raise the invalid argument on the empty list: they should return `0` and `0.` respectively. It would be nice to have this in our `v3` branch.

gasche:
Before I merge, could you indicate the PR number in the Changelog entry?

I have a remark related to that: it would be nice if you would keep a single patchset in a single PR, updating the PR with the new version of your patch, instead of closing the old PR and opening a new one as you currently do. This would avoid having `#TBD` in the Changes entry, and it would keep the whole discussion together in a single place instead of a linked list of PRs on each review. This is easy to do on your side: just use the same git branch for your patchset `list_favg` for example, and `--force-push` the new state of the branch to your personal clone. Github will automatically update the PR with the new patchset.

UnixJunkie:
I updated the PR number

gasche:
Thanks! Merged.

PR #747, opened 2017-04-11
preparatory work for next minor release (v2.6.0)

UnixJunkie:


gasche:
Wait, I think the next release should be 2.6, not 2.5.4. My mental model is that backward-incompatible changes bump the first number (major release), "just new features" bump the second number (minor release), and "just bugfixes" bump the third number (sometimes called "patch release"). See [Semantic Versioning](http://semver.org/).

gasche:
P.S.: Thanks for the release work!

UnixJunkie:
good catch

UnixJunkie:
I'm done with this. Another pair of eyes is welcome to check we did not overlook anything.

gasche:
I don't have the time to test this very thoroughly, but I had a quick look and it seems ok. Are you willing to push through the rest of the release work (that is: tag the release, and send the `opam` package), or would you rather have me announce it? I think at least that you should be the one announcing the release on the caml-list.

UnixJunkie:
I can continue tomorrow in the second half of the afternoon; I'll revert the opam file change (because I feel like you in fact)

PR #748, opened 2017-04-12
fix in the Makefile for when *.cmt{i} files are not produced by ocaml…

UnixJunkie:
…build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,9 +35,17 @@ endif
 
 INSTALL_FILES = _build/META _build/src/*.cma \
 	battop.ml _build/src/*.cmi _build/src/*.mli \
-	_build/src/*.cmti _build/src/*.cmt \
 	_build/src/batteriesHelp.cmo _build/src/batteriesConfig.cmo _build/src/batteriesPrint.cmo \
 	ocamlinit build/ocaml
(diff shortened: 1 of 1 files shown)


UnixJunkie:
That's for the release.
I'll have to retag if this fix is accepted.

gasche:
When does the "not produced" situation happen, don't we pass `-bin-annot` all the time? How did you notice the issue?

UnixJunkie:
we pass bin_annot in _flags, but old versions of ocamlbuild don't understand it (and just ignore it).
@yallop spotted that make install failed in the Travis CI tests for the new batteries opam package there: https://github.com/ocaml/opam-repository/pull/8942
Maybe there is a fix via myocamlbuild.ml but I don't want to venture in such lands (I'd rather like to change the build system altogether).

gasche:
Come on, the `myocamlbuild.ml` fix is trivial. I'll send a PR.

gasche:
But in fact the problem remains: `-bin-annot` was only added to 4.00 so if we want to support 3.12.1 (which is a good idea; Red Hat people still support it) we need to be robust to the absence of `.cmt*` files anyway.

gasche:
PR in #749. I think we should merge both.

PR #757, opened 2017-05-31
added BatArray.min_max

UnixJunkie:


gasche:
`BatList.min_max` takes a `?cmp:('a -> 'a -> int)` parameter to specify an alternative comparison function (but `.min` and `.max` don't, weird). Do we want to do the same thing here? I think that the choice you made is also reasonable (and the code seems perfectly mergeable).

UnixJunkie:
BatList.min_max is a much more recent addition than BatList.(min|max).
If we add a cmp optional parameter (defaulting to Pervasives.compare) that could be added
to all of them: min, max, min_max.

UnixJunkie:
I think this should be accepted as is (because it is coherent with the rest of the Array module).
I we want to add cmp functions to BatArray.(min|max|min_max), please open an issue (I might do it, but later).

PR #761, opened 2017-07-16
Fix typos in documentation strings

rixed:


gasche:
Thanks! The amount of typos is a bit disheartening, but I guess that is unavoidable.

Issue #756, opened 2017-05-31
batteries' code is not compatible with latest release of qcheck (0.6)


UnixJunkie:
you have to downgrade qcheck with opam (previous release) in order to pass 'make test'

    opam install qcheck.0.5.3.1

c-cube:
Sorry about the breakage. It's actually possible to modify the code for cheap (old `fun1` and `fun2` functions were renamed, not removed), or use the better new function generators.

PR #762, opened 2017-07-16
Fix make test

rixed:
Ask opam for a qcheck package that's compatible with batteries (This is not
a proper fix for #756, but make tests work in the meantime).
While at it, ask for bisect.
Also clean some harmless warnings in the compilation of some tests ("expression
should have type unit").


gasche:
Thanks! This looks fine. Medium-term we may be interested in migrating towards "bisect-ppx" that is more actively maintained, but keeping what we have working is nice.

Issue #760, opened 2017-07-08
Type signatures -- offering a consistent API for easy use


typesanitizer:
I'm primarily talking about the `BatVect` module below, but perhaps this point applies to other modules too (I haven't used them much).

Having the data structure be the final argument of the function is _really_ useful when using a chain of pipe operators; say you're working with a vector of vectors:
```ocaml
let x = [some computation]
        |> (fun v -> V.modify v p1 (fun z -> V.set z p2 y) 
```
versus a hypothetical
```ocaml
let x = [some computation]
        |> V.new_modify p1 (V.new_set p2 y) 
```
Obviously one can define such functions locally by rearranging the arguments, but I think that the library definitions should offer a consistent API where the last argument is always[1] the data structure for readability of long chains of pipe operations such as this.

Some functions in `BatVect` follow this, such as `append`, `concat` etc. but there are many which do not such as `modify`, `set` and so on.

For example, Haskell's [Data.List](https://hackage.haskell.org/package/base-4.9.1.0/docs/Data-List.html) follows this guideline and this point is explicitly mentioned in Elm's package [design guidelines](http://package.elm-lang.org/help/design-guidelines#the-data-structure-is-always-the-last-argument).

[1] For the specific case of `fold`-like operations, I'm not arguing one way or another (for the inner map). In this situation, Haskell and Elm are different. Haskell has
```haskell
foldl :: (b -> a -> b) -> b -> t a -> b 
foldr :: (a -> b -> b) -> b -> t a -> b
``` 
whereas Elm's `foldr` and `foldl` signatures are identical.
```elm
foldl : (a -> b -> b) -> b -> List a -> b
foldr : (a -> b -> b) -> b -> List a -> b
```
However, these functions still have the list as the last argument. I can understand that you might want to have the order of arguments be consistent with `Pervasives` and so you have the foldable type as the penultimate argument for `foldr`.

UnixJunkie:
The ocaml solution is to use labeled arguments. Look at BatList.Labels for example.
Then you can use the labeled parameters in whatever order you like.
I agree with you that "Having the data structure be the final argument of the function" is nice,
but it's a little bit late to enforce such a convention to the whole batteries library.

theindigamer:
> Look at BatList.Labels for example.

Thanks for the suggestion; it is a good alternative. However, BatVect (and many other modules) do not have such a module; are there any plans for adding them/would you accept a PR for it?

[...]

PR #763, opened 2017-07-17
Adds `Labels` submodule to `BatVect`.

theindigamer:
* Partial fix for issue #760.
* Also fixes a few typos in the `BatVect` documentation.

### DIFF
--- a/src/batVect.ml
+++ b/src/batVect.ml
@@ -595,12 +595,13 @@ let ord ord_val v1 v2 =
 module Labels =
 struct
   let init n ~f      = init n f
-  let concat ~v1 ~v2 = concat v1 v2
   let get v ~n       = get v n
   let at v ~n        = at v n
-  let set v ~n ~c    = set v n c
+  let set v ~n ~elem = set v n elem
   let modify v ~n ~f = modify v n f
   let sub v ~m ~n    = sub v m n
+  let insert ~n ~sub = insert n sub
+  let remove ~m ~n   = remove m n
   let iter ~f        = iter f
   let iteri ~f       = iteri f
   let map ~f         = map f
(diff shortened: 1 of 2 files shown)


theindigamer:
Before my commit, `make coverage` was already failing with 

```
File "qtest/all_tests.ml", line 3, characters 5-10:
Error: Unbound module OUnit
``` 
although `make test` worked just fine after downgrading to `qcheck.0.5.3.1`.

UnixJunkie:
Hello, thanks for the pull request.
Did you follow another module to choose the label names?
What is the meaning of ~c?
x0 is called init in BatList.Labels, for example.

theindigamer:
So I used `x0` for the folding accumulator as it was written [here](https://github.com/ocaml-batteries-team/batteries-included/wiki/Developers-guidelines) that you use `x0`. I had initially written it as `init` like BatList.Labels but changed it after reading the guidelines.

The `c` means character/cell, I suppose, as most of the existing documentation for BatVect uses `c` wherever a value of type `'a` is needed. For example, see the functions `make`, `append`, `prepend` and `set`. I also changed it to `c` in `destructive_set` for consistency. I suppose you could also use `e` for element everywhere instead of `c`.

Of course, this isn't a big deal so let me know what names are appropriate.

Also, is `make coverage` not to be used?

UnixJunkie:
OK, I am just annoyed by the x0 now: is any Label module using it? I see ~init in BatList and BatArray.
x0 doesn't say anything to me. Ok it is short but not very meaningful.

I never used make coverage. If it is broken, you could open another issue. make test was repaired very recently in another pull request that was accepted already.

gasche:
I agree that `x0` is not very nice. I would guess that the guidelines were written long ago and that we have not revisited this aspect. Using `~init` for consistency with List/Array seems reasonable to me, and you could update the guidelines in passing.

theindigamer:
Alright, I'll make those changes. Should I keep the `c` as is? Or use something else?

src/batVect.ml
```diff
@@ -1165,4 +1197,36 @@ struct
   let print ?(first="[|") ?(last="|]") ?(sep="; ") print_a out t =
     BatEnum.print ~first ~last ~sep print_a out (enum t)
 
+  module Labels =
+  struct
+    let init n ~f      = init n f
+    let concat ~v1 ~v2 = concat v1 v2
```
gasche: I believe that the labels are not useful nor very informative for `concat`; the spatial position of the arguments should tell us everything we need to know.

UnixJunkie:
I personally prefer 'e', standing for 'element' over 'c'.

src/batVect.ml
theindigamer: That's true but sometimes you might want to flip the arguments around. Say you created a vector using a chain of pipe operations and want to prepend it to another one. The labels are optional. I'll remove it if you don't want it.

theindigamer: Regarding leaving out functions, I've mostly included all but some like `insert` and `remove` were not included as the data structure is the last argument anyways and there doesn't seem much reason to use the labelled version (at least to me). Again, I'll put it in if you say so.

thizanne:
I believe the point of adding labels is to improve readability in exchange for a little bit of verbosity.  Thus why not use `~elem` rather than "`e` that stands for element" ? `f` is a common one-letter name for a function, `e` is not as much widespread. I'm all for being more explicit.

gasche:
I am ok with the current state. @UnixJunkie, @thizanne, is there a consensus to merge?

UnixJunkie:
I'm OK. Thanks to @theindigamer for having improved his patch.

theindigamer:
Cheers :smile:

thizanne:
Merged, thanks !

thizanne:
Sorry, I just read everyone was OK and went a bit fast. We should have a Changes entry and probably some tests -- I will do the work.

PR #767, opened 2017-07-30
Fix an implementation/specification mismatch in BatVect.insert

### LINKED ISSUE #766: Bug: off-by-one in BatVect.insert
Minimal example:

```
module V = BatVect;;
let a = V.of_list [0;1;2;3];;
V.get a 2 = 2;;
- : bool = true
V.(to_list @@ insert 3 (singleton 10) a);;
- : int list = [0; 1; 2; 10; 3]
```
According to the [documentation](https://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatVect.html), it should be inserted between indices 3 and 4, not 2 and 3.

gasche:
`BatVect.insert : int -> 'a t -> 'a t` is specified so that `insert n r u` inserts the rope `r` between the elements of index `n` and `n+1` in `u`. Its implementation does something different, it inserts `r` strictly before the element of index `n` in `u`.

The present commit changes the specification to match the implementation.

First, we expect users to have tested their program instead of trusting (or, unfortunately, even reading) the documentation, so it is likely that the uses of `BatVect.insert` with the current implementation is correct for their needs; changing it would break user program.

Second, we argue that the implemented behavior is actually far more
natural:

- the implementation (in terms of `sub` and `concat`) is nicer -- hear the code!

- the ranges of integer values valid for `insert` is  `0  .. (length u - 1)`, as one would expect, instead of  `(-1) .. (length u - 2)` as with the current specification.

- this gives a very natural invariant between `insert` and  `BatVect.remove : (*start*)int -> (*len*)int -> 'a t -> 'a t`:  `u` is equal to `u |> insert n r |> remove n (length r))`.

fixes #766.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,10 @@ Changelog
   #763
   (Varun Gandhi, review by Francois Berenger, Gabriel Scherer, Thibault Suzanne)
 
+- fix documentation of BatVect.insert to match (correct) implementation
+  #766, #767
+  (Gabriel Scherer, report by Varun Gandhi)
+
 - Documents exceptions for List.(min, max)
   #770
   (Varun Gandhi)
--- a/src/batVect.ml
+++ b/src/batVect.ml
@@ -318,8 +318,30 @@ let sub v s l = sub s l v
 let insert start rope r =
   concat (concat (sub r 0 start) rope) (sub r start (length r - start))
 
+(*$T insert
+(of_list [0;1;2;3] |> insert 0 (singleton 10) |> to_list) = [10;0;1;2;3]
+(of_list [0;1;2;3] |> insert 1 (singleton 10) |> to_list) = [0;10;1;2;3]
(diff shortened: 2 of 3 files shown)


UnixJunkie:
I get this when I do 'make test':

    File "src/batArray.ml", line 179, characters 40-51:
    Error: This expression has type int Q.arbitrary = int QCheck.arbitrary
           but an expression was expected of type
		    'a Q.Observable.t = 'a QCheck.Observable.t
I am using qtest.2.5.

c-cube:
You probably need to upgrade to qcheck 0.6, OR to update the test :)
The function generators have been updated (in the good direction, I like to think: no more magic, shrinking on functions, etc).
-- 
Simon

gasche:
Currently the tests in Batteries require `(qtest >= 2.0 & qtest < 2.5)`, see the opam file.

PR #768, opened 2017-07-30
Less exceptions -- remove internal uses of `BatReturn`

gasche:
@cpitclaudel reports that uses of `BatString` in the F* compiler become performance bottlenecks when used on a webpage through js_of_ocaml. This comes from the use of the BatReturn module that uses first-class exception declaration and raising for internal control flow -- this seems to be compiled to very slow js code.

This PR removes uses of local exception control-flow from the `String` and `Substring` modules. In passing, I found a bug in Substring.equal (when comparing substrings from different strings) that is fixed.

I have been working on removing these local-exceptions from all of Batteries, and have further PRs down the line (BatVect : easy; BatText : more challenging).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,7 +22,7 @@ Changelog
     raising/catching than the native OCaml backend.
     Internal exceptions (trough the BatReturn label) have been removed
     from the modules BatString and BatSubstring.
-  (Gabriel Scherer, request by Clément Pit-Claudel)
+  (Gabriel Scherer, request and review by Clément Pit-Claudel)
 
 - Documents exceptions for List.(min, max)
   #770
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,6 +15,15 @@ Changelog
   #766, #767
   (Gabriel Scherer, report by Varun Gandhi)
 
+- avoid using exceptions for internal control-flow
+  #768
+    This purely internal change should improve performances when using
+    js_of_ocaml, which generates much slower code for local exceptions
+    raising/catching than the native OCaml backend.
+    Internal exceptions (trough the BatReturn label) have been removed
+    from the modules BatString and BatSubstring.
+  (Gabriel Scherer, request by Clément Pit-Claudel)
+
 - Documents exceptions for List.(min, max)
   #770
   (Varun Gandhi)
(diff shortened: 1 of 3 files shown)


UnixJunkie:
While correcting the performance bottleneck of one user might be necessary, in the long run isn't it a problem of the js_of_ocaml compiler that doesn't know how to compile efficiently code using exceptions?
Many ocaml programmers rely on the fact that exception in ocaml are cheap.
With the recent introduction of local exceptions, this code style might even become more natural.

However, if the new code is not slower for most users (who don't run ocaml in a browser) and is easy to read, I am not against such PRs.

gasche:
Given that each modified function comes with a testsuite, I am reasonably confident that this patch preserves current behavior (except for the bug in batSubstirng.equal that is now fixed). I am thinking of merging the PR, even in absence of review, today.

cpitclaudel:
The code looks good to me :)
I have two small questions:

* In `starts_with`, could you pass just `i` to `loop` instead of `str`, `p`, and `i`, or is there an advantage to passing all three?
* In both implementations of find, could you declare `find` and `loop` as mutually recursive functions, instead of declaring `loop` inside of `find`? Would that be beneficial?

gasche:
Thanks for the review!

The "lifted" style gives closed functions which do not require closure allocation. When the number of parameters is small (smaller than 5, I think), this generally gives a (small) performance improvement. I don't usually care about it, but here I went for it on a whim.

I chose to use nested recursive functions to mirror the nested loops in the original implementation. Mutually recursive functions would of course also work, but they may make the structure of the control flow a bit less clear. (When the functions are not closed, mutually recursive functions are in fact a bit more efficient as their closure are allocated only once at function-declaration time, whereas a nested function function would be allocated at each turn of the loop; this does not matter here.)

cpitclaudel:
> The "lifted" style gives closed functions which do not require closure allocation. 

Super; this makes sense, thanks.

> I chose to use nested recursive functions to mirror the nested loops in the original implementation

And that too; thanks again :)

gasche:
I merged. @cpitclaudel, if you are still in review mood, #769 does the same change in `BatVect` function. (I have a patchset for the last remaining user of `BatLabel`, BatText, that I plan to send as a PR if #769 is merged.)

PR #769, opened 2017-07-30
Less exceptions in `batVect`

gasche:
This is a follow-up of #768, eliminating local exceptions for control flow from the BatVect module. As a side-effect, I added a `find_opt : ('a -> bool) -> 'a t -> 'a option` function to the existing `find : ('a -> bool) -> 'a t -> 'a` function.

### DIFF
--- a/src/batVect.mli
+++ b/src/batVect.mli
@@ -269,32 +269,39 @@ val exists : ('a -> bool) -> 'a t -> bool
     [ (p a0) || (p a1) || ... || (p an)]. *)
 
 val find : ('a -> bool) -> 'a t -> 'a
-(** [find p a] returns the first element of vect [a]
+(** [find p v] returns the first element of vect [v]
     that satisfies the predicate [p].
     @raise Not_found if there is no value that satisfies [p] in the
-    vect [a]. *)
+    vect [v]. *)
+
+val find_opt : ('a -> bool) -> 'a t -> 'a option
+(** [find_opt p v] returns [Some a], where [a] is the first element
+    of vect [v] that satisfies the predicate [p], or [None]
+    if no such element exists.
+
+    @since NEXT_RELEASE *)
 
 val mem : 'a -> 'a t -> bool
-(** [mem m a] is true if and only if [m] is equal to an element of [a]. *)
+(** [mem a v] is true if and only if [a] is equal to an element of [v]. *)
 
 val memq : 'a -> 'a t -> bool
 (** Same as {!Vect.mem} but uses physical equality instead of
     structural equality to compare vect elements.  *)
 
 val findi : ('a -> bool) -> 'a t -> int
-(** [findi p a] returns the index of the first element of vect [a]
+(** [findi p v] returns the index of the first element of vect [v]
     that satisfies the predicate [p].
     @raise Not_found if there is no value that satisfies [p] in the
-    vect [a].  *)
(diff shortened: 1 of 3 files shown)


gasche:
The diff currently contains all of #768. The idea is that once #768 is reviewed/merged I will rebase the current PR to the new trunk, to let reviewers look at only the batVect-related changes.

src/batVect.mli
```diff
@@ -269,32 +269,39 @@ val exists : ('a -> bool) -> 'a t -> bool
...
-    [f x] returns [true]. Operates in [O(n)] time. *)
+(** [filter f v] returns a vect with the elements [a] from [v] such that
+    [f a] returns [true]. Operates in [O(n)] time. *)
 
 val filter_map : ('a -> 'b option) -> 'a t -> 'b t
-(** [filter_map f e] returns a vect consisting of all elements
-    [x] such that [f y] returns [Some x] , where [y] is an element
-    of [e]. *)
+(** [filter_map f v] returns a vect consisting of all elements
+    [b] such that [f a] returns [Some x] , where [a] is an element
```
cpitclaudel: Should this be `Some b`?

src/batVect.mli
```diff
@@ -621,6 +628,13 @@ val find : ('a -> bool) -> 'a t -> 'a
     @raise Not_found if there is no value that satisfies [p] in the
     vect [a]. *)
 
+val find_opt : ('a -> bool) -> 'a t -> 'a option
+(** [find p a] returns [Some x] if [x] is the first element of vect [a]
```
cpitclaudel: I think this should say `find p v` instead of `find p a` and `a` instead of `x` for consistency with the other functions.

cpitclaudel: Also, this should say `find_opt`, not `find`

cpitclaudel: And in fact the phrasing of this `find_opt` is slightly different from the `find_opt` above, maybe it would be good to use the same text for both.

src/batVect.mli
```diff
@@ -269,32 +269,39 @@ val exists : ('a -> bool) -> 'a t -> bool
...
-(** [find p a] returns the first element of vect [a]
+(** [find p v] returns the first element of vect [v]
     that satisfies the predicate [p].
     @raise Not_found if there is no value that satisfies [p] in the
-    vect [a]. *)
+    vect [v]. *)
+
+val find_opt : ('a -> bool) -> 'a t -> 'a option
+(** [find_opt p v] returns [Some a], where [a] is the first element
+    of the vector [v] that satisfies the predicat [p], or [None]
```
cpitclaudel: Typo: `predicat`

gasche: Thanks, fixed.

src/batVect.mli
gasche: Indeed.

src/batVect.mli
gasche: I kept the old variable names here for consistency with the Make variable naming choice (I did not update it when I updated the variables around the normal `find` function; arguably this would be nicer, but la flemme.), but I fixed the other issues. thanks!

gasche:
Github auto-closed this based on my force-push activity, but this was not my intent. I will go ahead and merge. Thanks @cpitclaudel for the review -- I plan to send the `BatText` changes soon.

PR #770, opened 2017-07-31
Documents exceptions for List.(min, max) and small punctuation fixes.

theindigamer:
The exceptions were not documented earlier. Also, a few periods were missing.


UnixJunkie:
You don't need to fix the punctuations of all the documentation. However, documenting exceptions is useful.

theindigamer:
@UnixJunkie , it's just that I noticed some mistakes while reading so I fixed them. I didn't go out looking for punctuation mistakes.

PR #772, opened 2017-08-03
Add Changelog for #763 (BatVect.Labels addition)

### LINKED ISSUE #765: Add a Changes entry and tests for #763
I'll do that when I unbreak my ocaml installation.

thizanne:
Fix #765 , end of #763.

No test are already present in BatVect, I didnt add them for the Labels module. This should be fixed at some point, although @gasche suggests in #762 that we should change our tests system so maybe we should wait for this.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,7 +6,11 @@ Changelog
 - added BatArray.min_max
   #757
   (Francois Berenger)
-  
+
+- added a Label module to BatVect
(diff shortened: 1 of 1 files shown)


gasche:
Minor nitpick: Bisect is only the code coverage tool (this is what was discussed in #762). I have no plans to change how tests are written.

PR #773, opened 2017-08-09
Less exceptions battext

gasche:
This is the last PR of the "remove BatReturn" usage series after #768 and #769. After this there are no more uses in the Batteries sources themselves. (They may remain some usage of internal usage of exceptions for control-flow, though.)


gasche:
I am realizing only now that this PR is incomplete: it contains testsuites for the functions that use `BatReturn`, which is required to modify them later, but I have not included the Return-removing changes. I will prepare a complete patch and reopen the current PR when it is available. Sorry for the noise.

PR #777, opened 2017-08-11
Fix compilation with OCaml 4.05.0

### LINKED ISSUE #764: Compilation with OCaml 4.05.0
ocaml-batteries 2.6.0 fails to build with OCaml 4.05.0. I didn't see anything related to that in open issues, PRs or git log. I had to fix it myself in the Debian package (only the two last patches are relevant):

https://anonscm.debian.org/git/pkg-ocaml-maint/packages/ocaml-batteries.git/tree/debian/patches

I know, these patches are not totally satisfactory... but maybe someone can take inspiration from them to make a proper fix.

tchajed:
Fixes #764

### DIFF
--- a/src/batGc.mliv
+++ b/src/batGc.mliv
@@ -182,7 +182,8 @@ external counters : unit -> float * float * float = "caml_gc_counters"
     is as fast at [quick_stat]. *)
 
 ##V>=4.4##external minor_words : unit -> (float [@unboxed])
-##V>=4.4##  = "caml_gc_minor_words" "caml_gc_minor_words_unboxed" [@@noalloc]
+##V>=4.4##  = "caml_gc_minor_words" "caml_gc_minor_words_unboxed"
+##V<4.5##     [@@noalloc]
 ##V>=4.4##(** Number of words allocated in the minor heap since the program was
 ##V>=4.4##    started. This number is accurate in byte-code programs, but only an
 ##V>=4.4##    approximation in programs compiled to native code.
@@ -227,7 +228,8 @@ val allocated_bytes : unit -> float
     started.  It is returned as a [float] to avoid overflow problems
     with [int] on 32-bit machines. *)
 
-##V>=4.3##external get_minor_free : unit -> int = "caml_get_minor_free" [@@noalloc]
+##V>=4.3##external get_minor_free : unit -> int = "caml_get_minor_free"
+##V<4.5##          [@@noalloc]
 (** Return the current size of the free space inside the minor heap.
     @since 2.5.0 and OCaml 4.03.0 *)
 
--- a/src/batGc.mliv
+++ b/src/batGc.mliv
@@ -183,7 +183,7 @@ external counters : unit -> float * float * float = "caml_gc_counters"
 
 ##V>=4.4##external minor_words : unit -> (float [@unboxed])
 ##V>=4.4##  = "caml_gc_minor_words" "caml_gc_minor_words_unboxed"
-##V<4.5##     [@@noalloc]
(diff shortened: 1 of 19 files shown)


src/batInnerWeaktbl.ml
```diff
@@ -95,6 +95,7 @@ module Make (H: Hashtbl.HashedType) : Hashtbl.S with type key = H.t = struct
   let find_all tbl key =
     try all_value (W.find tbl (dummy key)) with Not_found-> []
   let find tbl key = top_value (W.find tbl (dummy key))
+  let find_opt tbl key = try Some (top_value (W.find tbl (dummy key))) with Not_found -> None
```
gasche: maybe you could write `try (Some (find tbl key)) ...` here? The code is a bit complex so I'd rather stay simple.

src/batGc.mliv
```diff
@@ -182,7 +182,8 @@ external counters : unit -> float * float * float = "caml_gc_counters"
     is as fast at [quick_stat]. *)
 
 ##V>=4.4##external minor_words : unit -> (float [@unboxed])
-##V>=4.4##  = "caml_gc_minor_words" "caml_gc_minor_words_unboxed" [@@noalloc]
+##V>=4.4##  = "caml_gc_minor_words" "caml_gc_minor_words_unboxed"
+##V<4.5##     [@@noalloc]
```
gasche: I'm afraid this will include "[@@noalloc]" on 4.03, which would be wrong. I think you need to use `##V=4.4##` instead.

src/batGc.mliv
```diff
@@ -227,7 +228,8 @@ val allocated_bytes : unit -> float
     started.  It is returned as a [float] to avoid overflow problems
     with [int] on 32-bit machines. *)
 
-##V>=4.3##external get_minor_free : unit -> int = "caml_get_minor_free" [@@noalloc]
+##V>=4.3##external get_minor_free : unit -> int = "caml_get_minor_free"
+##V<4.5##          [@@noalloc]
```
gasche: Here we need to have `[@@noalloc]` show up under both 4.03 and 4.04, but no other versions. Unfortunately, our prefilter code does not support disjunctions, so you may have to duplicate the line:

```
##V=4.3## [@@noalloc]
##V=4.4## [@@noalloc]
```

tchajed:
Thanks for your help @gasche! I'll go back and re-test everything under 4.03 and 4.04 when I get 4.05 working.

gasche:
Thanks. I was planning to work on this eventually, let me know if I can do something to help.

tchajed:
@gasche: the tests now pass in 4.03, 4.04, and 4.05. If things look good I'd like to rebase before merging.

gasche:
This looks nice, thanks, but I would like to ask for one more step: it would be nice if we could (re)implement some of the new library functions also in the older OCaml versions. This allows people using the new version of Batteries to write code that is backward-compatible with older OCaml versions. We don't do that for functions that are impossible to implement outside the compiler (eg new `Gc` features) or functions that are difficult to implement (I would recommend to leave Array0 out), but for example for all the `_opt` functions it is trivial to provide a naive implementation on older OCaml versions.

Let me know if that sounds like too much work for you, we could merge this PR and add this as separate work.

tchajed:
That makes sense. I could do this but not right away. Perhaps we should merge this PR and if nobody gets to adding backwards compatibility later I can do so.

For Array0 I realized there are some extra functions for `Array1`, `Array2` and `Array3` that batteries provides which we could also provide for `Array0`. They're just scalars so pretty much every function is trivial, but it provides some uniformity.

PR #778, opened 2017-08-11
Update for compatibility with qcheck 0.6

### LINKED ISSUE #756: batteries' code is not compatible with latest release of qcheck (0.6)

tchajed:
Unfortunately this does break backwards compatibility with qcheck 0.5,
so there is now an explicit opam dependency on qcheck >= 0.6.

Fixes #756.

### DIFF
--- a/opam
+++ b/opam
@@ -20,7 +20,8 @@ remove: [["ocamlfind" "remove" "batteries"]]
 depends: [
   "ocamlfind" {>= "1.5.3"}
   "ocamlbuild" {build}
-  "qtest" {test & >= "2.0.0" & < "2.5"}
+  "qtest" {test & >= "2.0.0"}
+  "qcheck" {test & >= "0.6"}
   "bisect" {test}
 ]
 available: [
--- a/opam
+++ b/opam
@@ -20,7 +20,7 @@ remove: [["ocamlfind" "remove" "batteries"]]
 depends: [
   "ocamlfind" {>= "1.5.3"}
   "ocamlbuild" {build}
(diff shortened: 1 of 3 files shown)


opam
```diff
@@ -20,7 +20,8 @@ remove: [["ocamlfind" "remove" "batteries"]]
 depends: [
   "ocamlfind" {>= "1.5.3"}
   "ocamlbuild" {build}
-  "qtest" {test & >= "2.0.0" & < "2.5"}
+  "qtest" {test & >= "2.0.0"}
+  "qcheck" {test & >= "0.6"}
```
gasche: I find it suspicious that the qtest versions before and after the patch do not have an empty intersection. We used the `< 2.5` bound to not do the backward-incompatible change, you suggest to move to the new version, shouldn't we now require `>= 2.5`? Are the versions between `2.0` and `2.5` really compatible with the new code?

tchajed: I didn't actually test this part of the change by having `opam` resolve dependencies.

You're right - `qtest < 2.5` are listed as conflicting with `qcheck` outright. I've inlined that constraint and just switched to `qtest >= 2.5`.

tchajed: Part of the confusion is that `qtest < 2.5` was somehow used to implement `qcheck <= 0.5`, which I don't completely understand.

gasche:
Oh, it is fully possible that the previous dependency specification was not very nice. Thanks for the update! Merging.

PR #779, opened 2017-08-13
Backport new 4.05 functions

gasche:
This is a followup from #777. Ideally I would like a nice version of this to be merged before our 4.05.0-supporting release -- hopefully this week or the next one.

For now only the *_opt functions are backported. I would like to backport `compare_lengths` and `compare_length_with` before merging. @tchajed seemed to think that `Bigarray.Array0` could also receive some easy work, but I haven't looked at it yet.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -41,6 +41,13 @@ Changelog
   #775
   (Gabriel Scherer)
 
+- Support for the new OCaml release 4.05
+  the `*_opt` functions and List.compare_lengths, compare_length_with
+  are also backported to older OCaml releases, so code using them from
+  Batteries should be backwards-compatible
+  #777, #779
+  (Tej Chajed, Gabriel Scherer)
+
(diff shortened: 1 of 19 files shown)


gasche:
@tchajed, something I forgot to check when I reviewed your 4.05 PR are the @since tags. "@since 4.05" does not really make sense for Batteries, since functions may be available for more OCaml versions than in the stdlib. To reference a future Batteries release, I use "NEXT_RELEASE" (easy to grep/sed at release time), and to reference an OCaml version we use "OCaml 4.05.0" (un-qualified version numbers are Batteries').

gasche:
I backported `compare_lengths` and `compare_length_with` from `List`. For `Array0`, we cannot implement the module itself as it requires a runtime change (I wasn't sure but I just checked), so as far as backporting is concerned there is nothing else to do. In particular, this PR is ready for review.

Issue #784, opened 2017-08-23
Revise API docs for newbies?


mars0i:
Hi,
I have two suggestions for the API docs pages to make them friendlier for newbies.   

The suggestion first is to add a note at the top that explains that, for example, `BatList` would normally be used as `Batteries.List`.  (In the long run it would better if it's described as `Batteries.List` or something simialr in the documentation, but this is a general issue with OCaml documentation that's been discussed online and that I imagine will eventually be resolved.  In the short run, a simple note would help new users to be a little bit less puzzled.) 

The second suggestion is to provide a page in which all of the modules are listed in alphabetical order, or listed in groups according to similar functions.  The current page http://ocaml-batteries-team.github.io/batteries-included/hdoc2 is no doubt useful for experienced users.  "New Data Structures" and "New Modules" are irrelevant to someone for whom everything is new, and most of the other categories on the page are not very useful to a new user, either.  If you're trying to find out what linear data structures are available, for example, you have to scan most of the page visually to find `BatList` and `BatArray`, or guess at their names and make the browser search the page.  I would find it easier to scroll to the "L"s if I thought there might be a List module.  (Actually, as it is, I know there's a `BatList` module, but I don't remember what section it's in, so I almost always type "list" into the browser search, which is less convenience than scrolling through an alphabet.)  It would be possible to have both an alphabetical page and the current page, maybe also with some sort of functionally categorized page.  However, I think the first page from the API link on the main web page should be to whatever page is easiest for newbies.  This can contain a link to the current API page for those who've been using Batteries for a long time.  That's my suggestion.

I can submit a PR for the first suggestion if that's desirable.  For the second suggestion, I don't now how the docs are generated, so maybe that should be left for someone else, if it seems desirable.

gasche:
The introduction to the API is in the file [build/intro.tex](https://github.com/ocaml-batteries-team/batteries-included/blob/master/build/intro.text). The syntax used is described in the [ocamldoc manual](http://caml.inria.fr/pub/docs/manual-ocaml-4.05/ocamldoc.html#sec333).

I'm a bit hesitant about the first change (some people use BatList directly and it's fine as far as I am concerned), but the second change definitely sounds like a potential improvement, so I would be curious to see what you propose.

mars0i:
[...]

PR #785, opened 2017-08-30
Add note explaining that BatX is also Batteries.X

mars0i:


UnixJunkie:
I guess this is only true after en 'open Batteries' directive

mars0i:
You mean before `open Batteries`?  Yes.  I took that to be implied by what I wrote.  I would be happy to write a bit more there.  The existing document is very brief at the top, so I was following that pattern.  A bit more explanation about `open` would be friendlier, though.  Leaving it implicit is less friendly, I would think, but means more text to scroll past to get to the modules.

mars0i:
Oh, forgot to mention that this for #784 .

mars0i:
I revised the statement to add a brief comment about what happens when `Batteries` has been `open`ed.  I meant for it to be a separate PR.  I forgot that you can't do that, so I guess I added a commit to the PR.  Anyway, this is a small PR, so it will be easy to edit if the Batteries team wants something different.

gasche:
You could give an example: "For example, `BatList` can also be used as `Batteries.List`."

gasche:
Merged, thanks!

PR #796, opened 2017-10-06
add BatMap.Big_int and BatSet.Big_int modules

matejkosik:
In the spirit of already existing (convenient) modules:
- `BatSet.Int`, `BatMap.Int`
- `BatSet.String`, `BatMap.String`
- ...

this PR adds two new similar modules:
- `BatSet.Big_int`
- `BatMap.Big_int`

### DIFF
--- a/src/batMap.ml
+++ b/src/batMap.ml
@@ -932,6 +932,7 @@ module Nativeint = Make (BatNativeint)
 module Float = Make (BatFloat)
 module Char = Make (BatChar)
 module String = Make (BatString)
+module Big_int = Make (BatBig_int)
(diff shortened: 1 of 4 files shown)


gasche:
Hi Matej! Num (which includes Big_int) is going out of the standard OCaml distribution in the next few releases, and is mostly deprecated in favor of `Zarith` these days, so I'm hesitant to add a dependency to it to some fairly central Batteries modules. (Batteries does depend on Num anyway, and this is probably not going away soon, but still.)

matejkosik:
OK. NB.

PR #798, opened 2017-10-23
added BatString.count_char

UnixJunkie:


UnixJunkie:
please consider merging this one before the new release

gasche:
I'm worried that the `fold` version is going to be sensibly slower than just a loop incrementing a reference.

UnixJunkie:
OK, I'll try to do some tests when I have time

UnixJunkie:
@c-cube Simon, any idea on this?

c-cube:
I don't think the `fold` should be slower, if the accumulator is just an unboxed integer, especially with flambda? But benchmarks are always good.

UnixJunkie:
```
open Core
open Core_bench

module S = struct

  include BatString

  let count_char_fold str char =
    fold_left (fun acc c ->
        if c = char then
          acc + 1
        else
          acc
      ) 0 str

  let count_char_while str char =
    let count = ref 0 in
    let n = length str in
    for i = 0 to n - 1 do
      if (unsafe_get str i) = char then
        incr count
    done;
    !count

end
    
let () =
  Command.run (Bench.make_command [
    Bench.Test.create ~name:"fold_ver" 
      (fun () -> S.count_char_fold "tototititata" 't');
    Bench.Test.create ~name:"while_ver" 
      (fun () -> S.count_char_while "tototititata" 't');
  ])
```

UnixJunkie:
```
# ./_build/default/benchit.exe
Estimated testing time 20s (2 benchmarks x 10s). Change using -quota SECS.
┌───────────┬──────────┬─────────┬────────────┐
│ Name      │ Time/Run │ mWd/Run │ Percentage │
├───────────┼──────────┼─────────┼────────────┤
│ fold_ver  │  51.10ns │  12.00w │    100.00% │
│ while_ver │  14.45ns │         │     28.28% │
└───────────┴──────────┴─────────┴────────────┘
```

UnixJunkie:
I updated the PR accordingly.
It should be merged by another maintainer.

UnixJunkie:
thanks

PR #799, opened 2017-10-23
added count_string

UnixJunkie:


UnixJunkie:
please consider merging this before the new release

gasche:
I forgot to ask for an entry in the ChangeLog.

I somewhat regret merging this PR in master, I think it would be cleaner to have a minor release with only the 4.06-related changes, even if that means making another minor release with your new functions shortly after that.

UnixJunkie:
It won't hurt anybody, don't regret.

PR #801, opened 2017-10-31
Revised docs for seq and from_loop, moved seq before from_loop

mars0i:
`seq`: Revised basic description to fit first part of `from_loop` doc,
which seems clearer.  Also added note that you can create an infinite
list by passing a function that constantly returns true.

`from_loop`: Gave additional detail about what the `next` function
is supposed to do, and noted that its result pair might contain
different types.  Added a simple example where the types are
the same.  (This example could be done more simpy using `seq`.
I still don't understand the purpose of `from_loop`.)

I moved `seq` before `from_loop` because it's a more basic version of this
kind of operation.  `from_loop` is more specialized, and complicated,
but its name might suggest that it's the normal way of making a lazy
list by repeated application of a function.  Users should see `seq`
first before trying to work out what `from_loop` is for.

### DIFF
--- a/src/batLazyList.mli
+++ b/src/batLazyList.mli
@@ -105,17 +105,22 @@ val from_while: (unit -> 'a option) -> 'a t
    results of [next].
    The list ends whenever [next] returns [None]. *)
 
+val seq: 'a -> ('a -> 'a) -> ('a -> bool) -> 'a t
+(**[seq init step cond] creates a lazy list from the successive results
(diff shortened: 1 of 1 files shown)


mars0i:
I'm going to submit a separate, improved PR from a new branch.  I recommend that this PR be closed.

mars0i:
(I'm surprised that I was allowed to close the PR.  I hope that's OK.)

PR #804, opened 2017-11-07
added BatSet.delta

UnixJunkie:


gasche:
> (** [delta l r] returns [(diff l r, inter l r, diff r l)] *)

Could you turn this property into a quickcheck test?

I'm also wondering about the performance of the implementation (your implementation does not allow copying sub-trees that only exist on either side). Would you compare the performance of your implementation against

```
    let delta l r = (diff l r, inter l r, diff r l)
```

on inputs of the form `delta set Empty` or `delta Empty set`? (For example with `set` having `10`, `1000`, and `100_000` elements.)

UnixJunkie:
OK, for this too I'll try to do some tests when I have time

UnixJunkie:
there was not much activity on this PR, so I guess the initial submitter (me) is not so much interested in the end

PR #806, opened 2017-11-13
Revised docs for seq, unfold, from_loop in LazyList

mars0i:
The start of the `seq` docs are now molded on start of the docs for `unfold` and `from_loop`, which I think was clearer.  Also added a note about how to create an infinite list.

`unfold`, `from_loop`: Added explanation of the signature of `next` and what it's for.  Added fibonacci example (which @gasche 

I also reordered the signatures so that the documentation would appear in an order that helps understanding a little bit.  `seq` is first because it's the simpler version, and `unfold` and `from_loop` are now adjacent since they do almost the same thing.  (`from_loop` is just a wrapper around `unfold`.)

Similar changes might be relevant for other modules such as `Enum`, but that can wait for another PR.

### DIFF
--- a/src/batLazyList.mli
+++ b/src/batLazyList.mli
@@ -106,26 +106,52 @@ val from_while: (unit -> 'a option) -> 'a t
    The list ends whenever [next] returns [None]. *)
 
 val seq: 'a -> ('a -> 'a) -> ('a -> bool) -> 'a t
-(**[seq init step cond] creates a lazy list from the successive results
+(**[seq data next cond] creates a lazy list from the successive results
    of applying [next] to [data], then to the result, etc.  The list
    continues until the condition [cond] fails.  E.g. [seq 1 ((+) 1) ((>) 100)]
    returns [[^1, 2, ... 99^]].  To create an infinite lazy list, pass 
    [(fun _ -> true)] as [cond].  If [cond init] is false, the result is empty. *)
 
+val unfold: 'b -> ('b -> ('a * 'b) option) -> 'a t
+(**[unfold data next] creates a (possibly infinite) lazy list from
+   the successive results of applying [next] to [data], then to the
+   result, etc. The list ends whenever [next] returns [None].  [next] 
+   should return an [option] of a pair whose first element will be the
+   current value of the sequence, and whose second element will be passed 
+   (lazily) to [next] in order to compute the following element.  One use 
+   is to allows each element of the resulting sequence to depend on the 
+   previous two elements, as in this definition of a Fibonacci sequence:
+
+   {[
+     let data = (1, 1)
+     let next (x, y) = Some (x, (y, x + y))
+     let fib = unfold data next
+   ]}
+
+   The first element [x] of the pair within [Some] will be the current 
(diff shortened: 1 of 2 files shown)


src/batLazyList.mli
```diff
@@ -105,22 +105,54 @@ val from_while: (unit -> 'a option) -> 'a t
...
+   [(fun _ -> true)] as [cond]. *)
 
 val unfold: 'b -> ('b -> ('a * 'b) option) -> 'a t
 (**[unfold data next] creates a (possibly infinite) lazy list from
    the successive results of applying [next] to [data], then to the
-   result, etc. The list ends whenever the function returns [None]*)
+   result, etc. The list ends whenever [next] returns [None].  The function 
+   [next] should return a pair [option] whose first element will be the
+   current value of the sequence; the second element will be passed 
+   (lazily) to [next] in order to compute the following element.  One use
```
gasche: One example use

src/batLazyList.mli
```diff
@@ -105,22 +105,54 @@ val from_while: (unit -> 'a option) -> 'a t
...
+val from_loop: 'b -> ('b -> ('a * 'b)) -> 'a t
+(**[from_loop data next] creates a (possibly infinite) lazy list from
+   the successive results of applying [next] to [data], then to the
+   result, etc.  The list ends whenever the function raises
+   {!LazyList.No_more_elements}.  The function [next] should return a pair 
+   whose first element will be the current value of the sequence; the second 
+   element will be passed (lazily) to [next] in order to compute the following
+   element.  One use of [unfold] is to make each element of the resulting sequence 
+   to depend on the previous two elements, as in this definition of a Fibonacci 
+   sequence:
```
gasche: This is very redundant with the documentation of `unfold` above. I think that you could stop after

> The list ends whenever the function raises {!LazyList.No_more_elements}.

and then just refer to the fact that otherwise it behaves as `unfold`.

mars0i:
In the `unfold` doc I added "example" before "use", but worded it a little bit differently.

For `from_loop`, I deleted most of the redundant material, as suggested, but added a note referencing `unfold` with a suggestion about what's different.  

I find `unfold` and `from_loop` initially difficult to understand.  I know that the `option`/no `option`' difference should be obvious from the signature, and is what one might expect given that `unfold` doesn't expect an exception while `from_loop` does.  However, for someone like me, at this stage, I think it's helpful to have the absence of `option` in `from_loop` made explicit because of the way in which these functions are already confusing without an additional twist.  That's the rationale for the last comment in the new `from_loop` doc.

Happy to go along with whatever seems best to others, though.

gasche:
Status report: I like the new state and would like to merge it, but the commit trail still has a terrible structure, so I need to do the work of cleaning it up (I'll publish a step-by-step log) before, and I've been drowned under other things to do. Ping me again if it's not done next week.

mars0i:
On this PR I didn't perform the `cherry_pick` operation you suggested on the other two PRs.  Happy to do that if you want.

mars0i:

gasche:
Thanks for this contribution!

PR #808, opened 2017-11-19
BatInt.Safe_int.mul performance improvements

murmour:
This patch makes the fast-path of `BatInt.Safe_int.mul` less expensive, by getting rid of the absolute value computation (which incurs branching). The downside is that the fast-path supports only non-negative integers from now on.

I am responsible for making the fast-path support negative numbers (#590). It made sense for a particular application of mine, but further experience showed that multiplication of positive integers is much more common in practice, in a wide variety of situations.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,9 @@ Changelog
 - add `BatString.count_string: string -> string -> int`
   #799
   (Francois Berenger)
+- Int: optimized implementation of Safe_int.mul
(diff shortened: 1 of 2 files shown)


gasche:
I looked at this and, while I don't understand the code very well as you do, I believe that the change is correct. Could you add a `ChangeLog` entry?

murmour:
Done.

PR #810, opened 2017-11-29
 Clarified documentation for `next` in LazyList

mars0i:


UnixJunkie:
So, seven commits just to change the ocamldoc for LazyList.next?

mars0i:
I had to undo some junk that was there because I had originally worked on master for a different PR.  I didn't want to just blow away my fork and start over, because I have another PR outstanding (which overwrote the junk).  My mess.  Sorry about that.

src/batLazyList.mli
```diff
@@ -287,7 +287,9 @@ val rindex_ofq : 'a -> 'a t -> int option
 *)
 
 val next : 'a t -> 'a node_t
-(**Compute and return the next value of the list*)
+(** Compute and return the first node from the list as a [Cons].  This 
+    differs from [hd], which returns first element (the first component of
```
gasche: which returns *the* first element

gasche:
Could you rebase your branch to remove the extra commits? One simple way to do it would be the following:

```
git checkout -b tmp # create a temporary branch in which to work
git cherry-pick 6b619cd # the hash of the commit actually doing the work in this branch
git checkout lazy_next_doc
git reset --hard tmp
git branch --delete tmp

mars0i:
Thanks very much for the git wizardry, @gasche.  

Did that do it?  Still looks a little noisy at my end.

mars0i:

gasche:
I will squash indeed as this is the least work-intensive way to resolve this PR, but I am a bit worried by the fact that, out of three different pull-requests, you were unable to offer a proper commit history. This is a useful skill to acquire, and it would be necessary for larger contributions where preserving the commit history is important (it contains several independent changes) -- which I hope you will have the occasion to contribute to Batteries. For this reason, my (implicit) offer to try to find out what's the issue and debug the commit history still stands -- or we may have this discussion again on your next PR.

mars0i:
Thanks @gasche.  

(I admit I am a bit puzzled about why git makes it so tricky to clean up a commit history.  My work habit is to make many little commits as I go along.  This is a safe strategy.  The problem is that no matter what branch this is done on, and what branch it is copied too, by default git wants to preserve every little bit of commit history.  I'm beginning to think that the best thing might be to deceive git by making a second repo on my machine that bears no relationship to the main parent repo, and then, when the code seems to be settled, copy it into a single commit in my main local copy of  the repo.  Or something like that.)

gasche:
In theory `rebase` is designed to help you transfer commits across different histories, see its documentation `git rebase --help` for more details.

mars0i:
@gasche, I'll study `rebase`--something I've been aware of but have avoided until now.

Issue #809, opened 2017-11-27
equal functions in Enum, Seq, and LazyList


mars0i:
`LazyList` lacks an `equal` predicate.  Both `Enum` and `Seq` include `equal`, but the syntax is different: `Enum` has a require equality-predicate parameter for comparison of individual elements, while `Seq` has an optional argument that defaults to `=`.  

I want to submit a PR for `equal` for LazyList, I'm not sure which model to follow.  I think the optional argument version of `Seq` makes more sense, since any of the usual standard OCaml types, except functions, will automatically work with `=`.   In fact I was considering submitting code more or less identical to what's in `Seq`, following kantian's suggestions in [this thread](https://discuss.ocaml.org/t/equality-of-lazylists/1192/14) in discuss.ocaml.org.

It would be nice if all three `equal` functions had the same syntax, but I gather that `Enum` is used a lot in Batteries, so changing `Enum`'s `equal` might break a lot of things.  I don't know how heavily `Seq` is used, and whether changing its syntax might break things.

Given the difference between `Enum`'s and `Seq`'s `equal` functions, I'm not sure how best to proceed.

gasche:
If you grep for `val equal` in the repository, you will see that having a mandatory argument is much more common than having an optional argument, so I would use that for consistency.

mars0i:
OK.  Should I "fix" `Seq`'s `equal`, too?  Or is that risky?

UnixJunkie:
@c-cube

gasche:
You shouldn't change the interface of an existing function unless we do a major release. You could send this change, separately, to the `v3` branch -- which we need to rebase on top of the current master, but that is another story.

mars0i:
If `equal` has a required `eq` argument, it's the same function as `for_all2`, except that it returns false rather than raising an exception if one list is longer and the test succeeds up to that point.

`Seq` and `Enum` don't have functions named `for_all2`, but `Enum`'s `equal` can also be used like `for_all2` if the enumerations are the same length.  You could do the same thing with `Seq.equal`; the function parameter defaults to `=`, but you could pass any function with the right signature.

I'm just a little befuddled, confused at the moment about what functions are needed.  An `equal` function with a required predicate argument is essentially `for_all2`, with `false` if the lists are different lengths.  Are both worth having?  I certainly like the idea of an `equal` function, and would write my own if it wasn't provided in `LazyList`.  

[...]

PR #811, opened 2017-11-29
Add lazylist equal predicate

mars0i:
This PR adds the function `equal` to `LazyList`.  It's a response to issue #809.

The function definition is essentially the one suggested by kantian on [discuss.ocaml.org](https://discuss.ocaml.org/t/equality-of-lazylists/1192/11); it's a slight modification of the existing definition of `for_all2`.

(The first six commits have nothing to do with equal; I mistakenly caused my copy of master to get out of sync, and I had to fix it.  Only the last three commits are relevant.)


gasche:
I like the patch, but not the shape of the patchset which is very noisy. You could use the method I proposed in https://github.com/ocaml-batteries-team/batteries-included/pull/810#issuecomment-348159639 to just cherry-pick the three last commits in a new branch starting from an up-to-date master.

mars0i:

src/batLazyList.ml
```diff
@@ -619,6 +619,14 @@ let for_all2 p l1 l2 =
...
   in aux l1 l2
 
+let equal eq l1 l2 =
+  let rec aux l1 l2 =
+    match (next l1, next l2) with
+    | (Cons (h1, t1), Cons(h2, t2)) -> eq h1 h2 && (aux t1 t2)
+    | (Nil, Nil)                    -> true
+    | (Cons _, Nil) | (Nil, Cons _) -> false
+  in aux l1 l2
+
```
gasche: As all new functions, this must come with unit tests. For the syntax of our inline tests, see the tests on `append` in the same module.

UnixJunkie:
Thanks for your contribution.

UnixJunkie:
@mars0i I updated the changelog file for you, please don't forget next time.
With more follow through, you could have had this merged way earlier.
Maybe just pinging in here after having added unit tests would have been enough.

PR #813, opened 2017-12-02
disable print_string_cap_ro/rw in battop to match #673

iitalics:
I was getting errors when starting up the REPL. This fixes it by commenting out lines that used functions disabled by dd4dbb4 .


gasche:
Thanks!

PR #814, opened 2017-12-04
Fix: incorrect behaviour of Map.union

murmour:
In case of conflicted bindings, `Map.union m1 m2` should prefer the value from `m2`, as stated in documentation.


thizanne:
Thanks !

PR #819, opened 2017-12-20
added BatSeq.{to_buffer|to_string|of_string}

UnixJunkie:


gasche:
`to_buffer` and `to_string` seem very reasonable to me and I would be in favor of merging. I have reservations about `of_string`; I find it strange that we would have a function that would accept `[1; 2; 3]` or `[1;2;3]`, but cannot be made to accept `[1; 2;3]` for example.

UnixJunkie:
The rational is that of_string expects the output of to_string.

UnixJunkie:
And, if you use something better than int_of_string (as I do in my tests), I guess your example would work.

gasche:
Another thing about the new function that I find unconvincing: the new exceptions. People usually use `failwith/invalid_arg` with a human-readable string as argument, or a new exception constructor to encode precisely the important information as data. You are using new constructors, but human-readable strings within. Would it not make sense to just use `failwith` instead? (In general I think that adding new ways to fail tend to make the interface more cumbersome to use.)

UnixJunkie:
I can split this into three separate pull requests.
It doesn't need to be accepted all at once.
I'm OK to failwith instead of using new exceptions.

gasche:
I would approve the version with `failwith`, or without `of_string`.

gasche:
P.S.: Please make sure if you use failwith that the string argument is informative enough (given that you lose the exception constructor).

UnixJunkie:
I updated the code.

src/batSeq.mli
```diff
@@ -285,9 +285,6 @@ val to_string : ?first:string -> ?last:string -> ?sep:string -> ('a -> string) -
     @since NEXT_RELEASE
 *)
 
-exception Wrong_prefix of string
-exception Wrong_suffix of string
```
gasche: I think the documentation comment also needs to be updated.

UnixJunkie:
done

UnixJunkie:
This PR has been updated several times and looks OK to me.
Since I am the original submitter, I don't like to merge my own PRs.
I think another maintainer should do it (which is a very sane safeguard regarding the integrity of our code base).

UnixJunkie:
thanks

PR #826, opened 2018-01-12
added BatArray.split

UnixJunkie:


gasche:
I just checked, this is about 20% faster on my machine than a simple two-pass implementation `let split a = (Array.map fst a, Array.map snd a)`. Merging.

PR #827, opened 2018-01-16
MultiPMap Documentation

sapristi:
Clarifies the use of some functions of the MultiPMap module.

### DIFF
--- a/src/batMultiPMap.mli
+++ b/src/batMultiPMap.mli
@@ -39,7 +39,7 @@ val is_empty : ('a, 'b) t -> bool
 (** returns true if the map is empty. *)
 
 val create : ('a -> 'a -> int) -> ('b -> 'b -> int) -> ('a, 'b) t
-(** creates a new empty map, using the provided function for key comparison.*)
+(** [create kcomp vcomp] creates a new empty map, using the kcomp for key comparison and vcomp for values comparison.*)
 
 val add : 'a -> 'b -> ('a, 'b) t -> ('a, 'b) t
 (** [add x y m] returns a map containing the same bindings as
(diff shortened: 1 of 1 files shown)


src/batMultiPMap.mli
```diff
@@ -33,13 +33,13 @@
...
 val empty : ('a, 'b) t
-(** The empty map, using [compare] as key comparison function. *)
+(** The empty map, using [compare] as comparison function for both keys and values. *)
 
 val is_empty : ('a, 'b) t -> bool
 (** returns true if the map is empty. *)
 
 val create : ('a -> 'a -> int) -> ('b -> 'b -> int) -> ('a, 'b) t
-(** creates a new empty map, using the provided function for key comparison.*)
+(** [create kcomp vcomp] creates a new empty map, using the kcomp for key comparison and vcomp for value comparison.*)
```
gasche: would you maybe cut this into two lines?

gasche:
This looks useful, thanks! Feel free to rebase into a single commit, otherwise I will squash and merge.

sapristi:
I'm not sure exactly how to rebase my commits, so I guess it would be easier for you to squash them.

gasche:
How to rebase commits is explained in the [Rewriting History](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) chapter of the [Git book](https://git-scm.com/book/), in particular the paragraph on "Squashing Commits" (but I would recommend reading the whole chapter).

(In the current case, we don't need to keep the commit messages of the various commits, so using `fixup` rather than `squash` (or editing the final commit message to remove the cruft) would be the best choice.)

If you intend keep contributing to git projects in the future (please do!), you should also try to follow the discipline of creating a new branch for each "contribution" (each general idea of change you want to submit upstream); working directly in `master` as you did here is not the best practice, for example because it makes it difficult for you to work on two contributions in parallel without mixing them. (You may think that this won't happen, but in general you will always regret not creating branches.)

sapristi:
Thanks I will look into this !

PR #828, opened 2018-01-18
simpler is better (and probably more efficient)

UnixJunkie:


UnixJunkie:
This is just simpler code, easy to decide if it should be merged or not.
But since I am the original author, I don't want to merge it myself, as a precaution.

gasche:
I would like to benchmark the difference first.

UnixJunkie:
I am too lazy to make a benchmark just for that.

PR #832, opened 2018-02-07
Compile Batteries with -no-alias-deps

rixed:
The size cost of using batteries has always been a problem. OCaml 4.02
introduced --no-deps-libs for this exact purpose (see
http://caml.inria.fr/pub/docs/manual-ocaml/extn.html#s%3Amodule-alias).

I couldn't find any discussion relative to this but I'm well aware this is
unlikely to not have been considered already.

Simple test:

```
  % cat toto1.ml
  open Batteries
  let () = Printf.printf "hello\n"
  % cat toto2.ml
  let () = BatPrintf.printf "hello\n"
  % for f in toto1 toto2 ; do \
      ocamlfind ocamlopt $f.ml -o $f.opt -package batteries -linkpkg ;
    done
  % ls -l toto*opt
```

Before the patch:

```
  -rwxr-xr-x 1 rixed 5272428 Feb  7 05:43 toto1.opt*
  -rwxr-xr-x 1 rixed 2006588 Feb  7 05:43 toto2.opt*
```

After:

```
  -rwxr-xr-x 1 rixed 3098452 Feb  7 05:38 toto1.opt*
  -rwxr-xr-x 1 rixed 2006588 Feb  7 05:38 toto2.opt*
```

That's a 40% reduction in size of a program using only BatPrintf from
Batteries, still bigger than a program using explicitly only BatPrintf
because of batPervasives, I assume.

There are several ways i can imagine this patch would break other programs:

- it might break the build of some programs relying on some of their direct
  dependencies to link in others of their direct dependencies, which I would
  qualify as a bug in those builds;

- this could prevent the dynamic linkage of some plugins if it expects the
  whole of batteries to be available from the main program;

- programs relying on any side effects of some otherwise unused batteries
  module initialization code, but that sounds very unlikely.

Note that I'm not familiar with this option so there might be others, more
obvious issues. I'm confident you will know :-)

Note: I have no idea if/how the "pack" option that seems to be present in the
setup.ml could be used for the same purpose.

### DIFF
--- a/_tags
+++ b/_tags
@@ -11,4 +11,5 @@ true: package(bytes), warn_-3, bin_annot
 "examples": -traverse
 <src/batOpaqueInnerSys.*>: opaque
 true: safe_string
+true: no_alias_deps
 


UnixJunkie:
I never do open Batteries, one more reason to continue doing so.
The proposed patch looks interesting.

gasche:
That's a reasonable idea. The main problem that I see is your comment on dynamic linking, but I am not sure that a problem would arise there and I think that we could give this a try anyway. Two comments on the implementation:

- have you tried just adding `true: no_alias_deps` to `_tags` first, and see if that works? In theory the flag should just be ignored on non-supporting OCaml versions, so you wouldn't need the version-conditional-logic
- the way you implement the feature flag, it is really a feature test, rather than an ocanl-version test, so your Makefile variable could be `$OCAML_HAS_NOALIASDEPS` for clarity.

rixed:
Indeed adding the option into `_tags` does the trick.
When a flag is not known a warning is emmitte though:

```
Warning: the tag XXX is not used in any flag or dependency declaration, so it will have no effect; it may be a typo. Otherwise you can use `mark_tag_used` in your myocamlbuild.ml to disable this warning.
```

A quick look at that myocamlbuild.ml file didn't reveal where/how to use that and the structure of that file was not clear to me. A Google search only revealed that others have fought this in the dark. All in all, this innocuous warning on old pre-4.02 versions looks acceptable to me. It's less likely to scare a new user than the warning about the missing package 'bisect', FWIW.

So updating the PR.

rixed:
So what's the status regarding this? Are we allowing more time for others to intervene?

UnixJunkie:
I think we should merge this.

UnixJunkie:
@gascher not decided yet? ;)

PR #833, opened 2018-02-15
A couple of critical fixes in Map.update

murmour:
Two separate issues are fixed by two separate commits, which should be self-explanatory.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -19,6 +19,9 @@ Changelog
   prefer the value from [m2], as stated in documentation.
   #814
   (Max Mouratov)
+- Fix: [Map.update k1 k2 v m] did not work correctly when [k1 = k2].
+  #833
+  (Max Mouratov)
 
 ## v2.8.0 (minor release)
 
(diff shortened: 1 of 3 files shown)


UnixJunkie:
wasn't there a unit test supposed to catch this already?
Is the test broken / never launched?

murmour:
Tests were not comprehensive enough.

Also, I suppose that no one has _ever_ used this function.

UnixJunkie:
It was probably added to reflect BatSet.update (by me).
Nice catch.
I am in favour of merging this, unless @gasche has some objections.

UnixJunkie:
Can you send a fix for BatSet.update too?
I think it suffers from the same problem.

murmour:
Indeed, BatSet.update suffers from one of these issues. I have added a fix for that.

UnixJunkie:
I confirm the new tests are passing.
Thanks for the fix in BatSet also.

gasche:
Thanks!

PR #836, opened 2018-02-19
BatString.ml should use unsafe versions of set and get

murmour:
This gives a slight performance boost to one of my programs.

### DIFF
--- a/src/batString.mlv
+++ b/src/batString.mlv
@@ -398,6 +398,7 @@ let split_on_char sep str =
     (* str is non empty *)
     let rec loop acc ofs limit =
       if ofs < 0 then sub str 0 limit :: acc
+      (* ofs >= 0 && ofs < length str *)
       else if unsafe_get str ofs <> sep then loop acc (ofs - 1) limit
       else loop (sub str (ofs + 1) (limit - ofs - 1) :: acc) (ofs - 1) ofs
     in
@@ -652,6 +653,7 @@ let fold_lefti f init str =
   let n = String.length str in
   let rec loop i result =
     if i = n then result
+    (* i >= 0 && i < len str *)
     else loop (i + 1) (f result i (unsafe_get str i))
   in loop 0 init
 (*$T fold_lefti
@@ -665,7 +667,9 @@ let fold_right f str init =
   let rec loop i result =
     if i = 0 then result
     else
+      (* i > 0 && i <= len str *)
       let i' = i - 1 in
+      (* i' >= 0 && i' < len str *)
       loop i' (f (unsafe_get str i') result)
   in
   loop n init
@@ -679,7 +683,9 @@ let fold_righti f str init =
   let rec loop i result =
     if i = 0 then result
     else
+      (* i > 0 && i <= len str *)
       let i' = i - 1 in
+      (* i' >= 0 && i' < len str *)
       loop i' (f i' (unsafe_get str i') result)
   in loop n init
 (*$T fold_righti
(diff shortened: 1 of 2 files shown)


UnixJunkie:
Is it always guaranteed that i is in [0..n-1]?
This will need to be carefully reviewed by several maintainers I think.

murmour:
> Is it always guaranteed that i is in [0..n-1]?

I have carefully reviewed the changes again, and believe so, yes.

gasche:
In 266d5f6ca506ea779788db5cb7d6fd2f2521a646 I started annotating occurrences of `unsafe_{get,set}` with static assertions in comments, to check that the invariants are indeed respected. I think we should consider completing this annotation work before merging the PR, but I have other things to do for the rest of the day. Feel free to help!

(In OCaml, `assert` is not disabled in builds, so using `assert` for static conditions is not a good way to remove array-access checks. It would be nice if we could use something like Why3 from OCaml to check those facts statically.)

UnixJunkie:
I'll try to read the full diff tomorrow.
I need a fresh head.
I am worried it could insert security problems.

murmour:
@gasche 
> In 266d5f6 I started annotating occurrences of unsafe_{get,set} with static assertions in comments, to check that the invariants are indeed respected. I think we should consider completing this annotation work before merging the PR, but I have other things to do for the rest of the day. Feel free to help!

A great idea!

I've added assertions to those functions that are affected by the PR (except a couple of really trivial cases), so that we could be more sure that the change does not cause regressions.

murmour:
@gasche 
> In 266d5f6 I started annotating occurrences of unsafe_{get,set} with static assertions in comments, to check that the invariants are indeed respected. I think we should consider completing this annotation work before merging the PR, but I have other things to do for the rest of the day. Feel free to help!

A great idea!

I've added assertions to those functions that are affected by the PR (except a couple of really trivial cases), so that we could be more sure that the change does not cause regressions.

src/batString.mlv
```diff
@@ -934,14 +953,15 @@ let numeric_compare s1 s2 =
   BatEnum.compare (fun g1 g2 ->
     let s1 = of_enum g1 in
     let s2 = of_enum g2 in
-    if BatChar.is_digit (unsafe_get s1 0) && BatChar.is_digit (unsafe_get s2 0) then
+    (* groups are never empty, so index 0 is always valid *)
+    if BatChar.is_digit (unsafe_get s1 0) &&
+       BatChar.is_digit (unsafe_get s2 0) then
```
gasche: I don't trust the `Enum` module enough to ensure this property (or have you went to annotate all `Enum` functions involved here with static assertions to "prove" this as well?). You don't risk undefined behavior on trusting another module's interface specifications. I that you must use safe accessors here -- in any case, the overhead of Enum largely drowns out any bound access cost.

murmour: Your arguments are convincing. I've reverted this change.

UnixJunkie:
'len-1' should be written 'len - 1'.
There are several changes like this, this is not the standard OCaml convention.

UnixJunkie:
Appart for lines 405..410 and 732 (indentation), I am OK with the changes.
It would have been easier to review if only unsafe_get was introduced, not unsage_get plus
some indentation changes.
Sorry for sounding grumpy, but this is a useful contribution so thank you.

murmour:
I've reverted a few unnecessary changes.

murmour:
The branch was rebased to resolve conflicts. Are there any outstanding issues remaining?

gasche:
Thanks for the ping, merged.

(For the record, I have mixed feelings about using unsafe operations, but I believe we have used due diligence here to avoid problems.)

PR #841, opened 2018-02-26
A couple of improvements to Array.insert

murmour:
Self-explanatory changes are separated into two commits.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -35,6 +35,9 @@ Changelog
   The assertion is now documented.
   #841
   (Max Mouratov)
+- Implementation of Array.insert now uses [unsafe_get] and [unsafe_set].
+  #841
+  (Max Mouratov)
 
 ## v2.8.0 (minor release)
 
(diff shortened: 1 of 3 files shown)


gasche:
The Change entry suggests that you print a refine message on *negative indices*, but that is not the case, you refined the general message to point out that the failing function is `Array.insert` (there is no specific message on underflows as opposed to overflows). Otherwise the patch is correct and an improvement. Should we merge as-is, or do you wish to clarify the change entry?

murmour:
> The Change entry suggests that you print a refine message on negative indices, but that is not the case, you refined the general message to point out that the failing function is Array.insert.

Got it. I agree that the wording could be more precise.

I've changed "negative" to "invalid".

gasche:
Merged, thanks.

UnixJunkie:
the valid range is [0..n], not [0..n-1]?

gasche:
This confused me at first, but yet. The returned array has size (n+1), and inserting at position n is supported (it puts the element at the end of the new array). The code is correct and this use-case is tested. It may have been emphasized by the documentation, but I didn't want to add unrelated concerns to the PR.

UnixJunkie:
OK then, it's an insert that can append at the end of the array, sorry for the noise.

PR #843, opened 2018-03-01
Do not depend upon bisect any longer

rixed:
Rational: bisect is not very actively maintained any more and code coverage
have not been useful (or is not useful any more).


gasche:
Merged, thanks!

PR #847, opened 2018-03-10
Less magical DynArray

murmour:
It always bothered me that DynArray makes so much use of undocumented compiler internals, including shady hacks to handle arrays of floats. This PR is an attempt to make the implementation more simple, without introducing performance regressions. 

My changes are based on the following knowledge:
* `%obj_size` is an alias for `%array_length` (Array.length);
* `%obj_field` is an alias for `%array_unsafe_get` (Array.unsafe_get);
* `%obj_set_field` is an alias for `%array_unsafe_set` (Array.unsafe_set);
* `%caml_obj_dup` (Obj.dup) and `%caml_array_sub` (Array.copy) are similar in performance + the latter handles float arrays correctly (judging by implementation);
* `%caml_obj_block` (Obj.new_block) and `%caml_make_vect` (Array.make) are similar in performance + the latter allocates float arrays when needed (judging by implementation);
* Copying sub-arrays with `%caml_array_sub` should be faster than using a loop.

The only magical piece left is this:
```ocaml
(* A polymorphic placeholder that is never accessed by the user *)
let dummy_for_gc = Obj.magic 0
```

I don't think it's possible to get rid of it without asking the user to provide a default value, making the interface ugly and backward-incompatible, so we probably have to live with it, for the time being.

Regarding performance: as expected, there are no measurable regressions in a couple of my programs that make heavy use of DynArray. I don't think a dedicated benchmark is necessary.

Edit: grammar.


murmour:
> I don't think it's possible to get rid of it without asking the user to provide a default value

A sidenote: in my personal stdlib extension many modules export a `zero` value, which is typically used as a placeholder in various data structures that are parametrized by said modules. Perhaps, something similar is needed in Batteries? For example, there is no `BatString.zero` (which could be an empty string).

gasche:
I haven't looked at the patch yet, but I appreciate the move, thanks. (Don't look at us accusingly, the Dynarray code comes from Extlib!)

One small remark: when Flambda is used, it is important to use `Sys.opaque_identity` to prevent the optimizer from reasoning about weird values -- here it should wrap the `Obj.magic` in the `dummy_for_gc` definition. However, it is only available since 4.03.0, so to use it we have to move DynArray to a `.mlv` for conditional compilation.

murmour:
> Don't look at us accusingly, the Dynarray code comes from Extlib!

You are not the first person to point out the accusing look of monsieur Pierre Bonnard on my avatar. I should probably change it to something more friendly (like a cat or a pokémon) :-)

murmour:
> One small remark: when Flambda is used, it is important to use Sys.opaque_identity to prevent the optimizer from reasoning about weird values -- here it should wrap the Obj.magic in the dummy_for_gc definition. However, it is only available since 4.03.0, so to use it we have to move DynArray to a .mlv for conditional compilation.

Done.

gasche:
Let me share a meditation about the interaction with the float-array optimization. The previous (unsafe) code had the property that a regular array-like representation was always used for storage.

You mention that using arrays (`make_vect` and `sub`) enables the float-array optimization, but if I understand correctly this optimization is *incorrect* in presence of an integer dummy value (for example, writing the integer dummy value into an (unboxed) float-array would segfault).

If that analysis is correct, then we must ensure that the new code does *not* use the array optimization. And it is not clear to me how to check that, and in fact I am a bit worried of the correctness of your patch as a result. I am reasonably confident that all fresh `arr`-field creations go through `imake`, which passes the dummy_gc_value initializer and will thus never be a float-array. But if `Array.copy` and `Array.sub` would ever re-pack a generic-array containing floats into an unboxed float-array, then we would be in trouble with your patch.

gasche:
After going to read more code in `byterun/array.c`, I notice that the OCaml runtime assumes that *all* `float array` values have tag Double_array_tag. The errors that I was worried about above does not occur, but your PR can actually be used to break this invariant, which is indeed unsound:

From `batteries-included/_build/src` from a build of your branch:

```
# #use "topfind";;
# #require "num";;
# #load_rec "batDynArray.cmo";;
# let danger = BatDynArray.(to_array (init 1 float_of_int));;
val danger : float array = [|0.|]
# let real_float_array = [|0.|];;
val real_float_array : float array = [|0.|]
# Array.blit real_float_array 0 danger 0 1;;
- : unit = ()
# danger;;
rlwrap: warning: ocaml crashed, killed by SIGSEGV (core dumped).
```

gasche:
Note: the current `of_array` implementation (the one you are replacing) is doing the right thing, with a slow path when the input array would have `double_array_tag`. You could represerve this in your new code, and make `to_array` safe using the same trick: in the non-empty case, check the tag of the first dynarray element, if it is `double_tag` then you need to copy element-by-element for safety, otherwise using `isub` is a correct fast path. I believe that this would suffice to make your implementation sound, but I have not checked carefully.

(The invariant to preserve is that internal `array` values must never have the double_array_tag representation, and that a float-array without double_array_tag must never escape to the outside world.)

I am not sure that the proposed change, even after it has been made sound, is worth it. Instead of low-level allocation code, we use a higher-level container in an unsafe way, and we have to preserve invariants that are exactly the same as what the low-level version had to keep in mind in the first place.

murmour:
_... crawls slowly out of a deep rabbit hole, holding a can of worms ..._

Wow! Crazy stuff! Thanks, Gabriel, for sharing your meditation.

Your analysis is correct: `of_array` now returns an invalid array of floats; and monkey-patching the issue by replicating old hacks makes the whole change very questionable.

I've been meditating myself, and decided that the the best (and only) way to resolve this surprising issue is to get rid of `Obj.magic` totally. One approach:

```ocaml

type 'a t =
  {
    mutable arr: 'a array;
    mutable len: int
    init_size: int
  }


let make init_size =
  {
    len = 0;
    arr = [||];
    init_size;
  }

let add d v =
  begin match Array.length d.arr with
    | 0 ->
        d.arr <- Array.make d.init_size v;
    | n when n = d.len ->
        let new_arr = Array.make (n*2) v in
        Array.blit d.arr 0 new_arr 0 n;
        d.arr <- new_arr;
    | _ ->
        Array.unsafe_set d.arr d.len v;
  end;
  d.len <- d.len + 1
```

Pros: zero magic; unboxed float arrays; compatible with Flambda.
Cons: `init_size` takes additional memory;  `delete` becomes leaky, since we can't "zero" a value.

In my opinion, pros do outweigh the cons, if just slightly.

gasche:
A middle ground would be to use the magic only on deletion:

```
let dummy_float = 0.
let dummy_for_gc (type a) (v : a) : a =
  Sys.opaque_identity @@
    if Obj.tag (Obj.repr v) = Obj.double_tag
    then Obj.magic dummy_float
    else 0

let delete =
  ...
  let dummy = dummy_for_gc (iget d.arr (d.len - 1)) in
  iset d.arr (d.len - 1) dummy
 ...
```

If you wanted to go all-safe, maybe `'a option array` could be an option as it is not leaky.

gasche:
Or

```
let erase arr n =
  if Obj.tag (Obj.repr arr) = Obj.double_array_tag then ()
  else arr.(n) <- Sys.opaque_identity (Obj.magic 0)
```

murmour:
> If you wanted to go all-safe, maybe 'a option array could be an option as it is not leaky.

For many use cases, the hit on performance would be unacceptable.

> A middle ground would be to use the magic only on deletion.

I like this option -- restricting magic to just 3 functions: `delete`, `delete_last`, and `delete_range`. If you think that the overall plan is OK, I will find time to rework DynArray (it would even feel good for the soul).

gasche:
Yes, I tihnk the overall plan is nice. Thanks for this work!

UnixJunkie:
I understand this can be closed, and @murmour will send another PR in the future.

Issue #844, opened 2018-03-02
Fix String documentation


thizanne:
BatString documentation should be taken care of:  there is no organisation, some functions are badly documented (`right` especially refers to `left`), and so on. 

This is a big and ungrateful work, so let's just leave this issue open as a reminder. Maybe other modules could also benefit from some work, but `String` appeared as quite neglected as I was trying to find a function in the documentation.

fccm:
The ocamldoc of String doesn't look that bad to me (I don't really see what should be improved).
This comment was written in 2018, so was it improved since then?
Only the functions that was switched to use the new `bytes` type should maybe have a deprecated tag added so that they could be removed in the future.

PR #849, opened 2018-03-11
Fix docstring of String.right

murmour:
Reported by @thizanne in #844.


UnixJunkie:
Thanks, can you update the changelog too?

murmour:
I don't think it was worth an entry, but alright: done.

PR #850, opened 2018-03-17
A clean-up of invalid_arg uses

murmour:
This patch makes uses of `invalid_arg` more uniform, fixing the following kinds of irregularities:
* Lack of module name in a message (e.g. `Heap.find_min` threw "find_min");
* Lack of function name in a message (e.g. `List.reduce` threw "Empty list");
* Wrong module name in a message (e.g. `PathGen` functions threw messages with "Path");
* Wrong function name in a message (e.g. `List.fold_left2` threw a message with "fold_right2");
* Minor style issues: capitalization, punctuation, "Bat" prefixes;
* Use of `raise (Invalid_argument "...")` instead of `invalid_arg "..."`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -34,7 +34,7 @@ Changelog
   raised exceptions
   #840, #754
   (FkHina)
-- Fix: [Array.insert] now throws a more relevant message on invalid indices
+- Fix: [Array.insert] should throw a more relevant message on invalid indices
   instead of the generic [invalid_arg "index out of bounds].
   The assertion is now documented.
   #841
@@ -45,6 +45,13 @@ Changelog
 - Fix documentation of [String.right].
   #849, #844
   (Max Mouratov, reported by Thibault Suzanne)
+- Fix: [Heap.del_min] should throw [Invalid_argument] with the specified
+  "del_min" message instead of "find_min_tree".
+  #850
+  (Max Mouratov)
+- More uniform and correct [Invalid_argument] messages.
+  #850
+  (Max Mouratov)
 
 ## v2.8.0 (minor release)
 
--- a/src/batBitSet.ml
+++ b/src/batBitSet.ml
@@ -48,7 +48,7 @@ let capacity t = (Bytes.length !t) * 8
 let empty () = ref (Bytes.create 0)
 
 let create_ sfun c n = (* n is in bits *)
-  if n < 0 then invalid_arg ("BitSet."^sfun^": negative size");
+  if n < 0 then invalid_arg ("BitSet." ^ sfun ^ ": negative size");
   let size = n / 8 + (if n mod 8 = 0 then 0 else 1) in
   ref (Bytes.make size c)
 
@@ -71,7 +71,7 @@ type bit_op =
 let rec apply_bit_op sfun op t x =
   let pos = x / 8 in
   if pos < 0 then
-    invalid_arg ("BitSet."^sfun^": negative index")
+    invalid_arg ("BitSet." ^ sfun ^ ": negative index")
   else if pos < Bytes.length !t then
     let delta = x mod 8 in
     let c = Char.code (Bytes.unsafe_get !t pos) in
@@ -105,7 +105,7 @@ let toggle t x = apply_bit_op "toggle" Toggle t x
 let mem t x =
   let pos = x / 8 in
   if pos < 0 then
-    invalid_arg ("BitSet.mem: negative index")
+    invalid_arg "BitSet.mem: negative index"
   else if pos < Bytes.length !t then
     let delta = x mod 8 in
     let c = Char.code (Bytes.unsafe_get !t pos) in
--- a/src/batFile.ml
+++ b/src/batFile.ml
@@ -45,8 +45,8 @@ let perm l =
     ~f:(fun acc x -> acc lor x)
 
 let unix_perm i =
-  if 0<= i && i <= 511 then i
-  else raise (Invalid_argument (Printf.sprintf "Unix permission %o " i))
+  if 0 <= i && i <= 511 then i
+  else Printf.ksprintf invalid_arg "File.unix_perm: Unix permission %o" i
 
 (* Opening *)
 type open_in_flag =
--- a/src/batHeap.ml
+++ b/src/batHeap.ml
@@ -108,7 +108,7 @@ let find_min bh = match bh.mind with
 
 
 let rec find_min_tree ts k = match ts with
-  | [] -> failwith "find_min_tree"
+  | [] -> invalid_arg "del_min"
   | [t] -> k t
   | t :: ts ->
     find_min_tree ts begin
@@ -285,7 +285,7 @@ module Make (Ord : BatInterfaces.OrderedType) = struct
     | Some d -> d
 
   let rec find_min_tree ts k = match ts with
-    | [] -> failwith "find_min_tree"
+    | [] -> invalid_arg "del_min"
     | [t] -> k t
     | t :: ts ->
       find_min_tree ts begin
--- a/src/batHeap.ml
+++ b/src/batHeap.ml
@@ -107,35 +107,44 @@ let find_min bh = match bh.mind with
 *)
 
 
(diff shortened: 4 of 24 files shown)


UnixJunkie:
find_min_tree became del_min in the exception message.

UnixJunkie:
del_min_tree is called del_min

murmour:
> find_min_tree became del_min in the exception message.

Yep, because `Heap.del_min` is the only caller of the internal `find_min_tree` function.

murmour:
> del_min_tree is called del_min

Its only caller is also `Heap.del_min`.

ChangeLog
```diff
@@ -45,6 +45,11 @@ Changelog
 - Fix documentation of [String.right].
   #849, #844
   (Max Mouratov, reported by Thibault Suzanne)
+- More uniform and correct Invalid_argument messages.
+  #850
+  (Max Mouratov)
+
+>>>>>>> Add ChangeLog entry
```
murmour: This is an accidental rebase artefact. Will fix it after a review.

UnixJunkie:
Those internal functions (if they are not declared in the interface) should be sub functions of del_min then.

UnixJunkie:
I'm OK with the changes.
However, they might break existing code if users where pattern-matching on the exception string/message.
This is a bad practice but may happen.
Depending on some other maintainer's opinion, this might be merged.
Thanks for the very many small fixes, they do increase the quality of the code.

murmour:
> Those internal functions (if they are not declared in the interface) should be sub functions of del_min then.

Probably, yes, if only to make it clear to the reader that error messages are correct. Shall I refactor this?

UnixJunkie:
Yes, that would be my suggestion.

murmour:
> However, they might break existing code if users where pattern-matching on the exception string/message. This is a bad practice but may happen.

Wow: I just noticed that error messages of `Heap.find_min` and `Heap_del_min` are actually specified in docstrings:
```ocaml
val find_min : 'a t -> 'a
(** Find the minimal element of the heap. O(1)
    @raise Invalid_argument ["find_min"] if the heap is empty *)

val del_min : 'a t -> 'a t
(** Delete the minimal element of the heap. O(log n)
    @raise Invalid_argument ["del_min"] if the heap is empty *)
```

I'll have to revert my changes to these functions. These are the only functions in whole Batteries for which the message is actually specified.

I'm not so worried about changing other functions, because expecting the text of unspecified error messages to be constant is an extremely bad practice indeed, and we should expect our users to be better.

murmour:
Last update changes:
* `Heap.find_min` now throws "find_min", as before (a specified behaviour);
* `Heap.del_min` now throws the specified "del_min" message instead of "find_min_tree". I think this qualifies as a bugfix.
* Implementation of `Heap.del_min` is refactored according to @UnixJunkie suggestions (in a separate commit).

gasche:
We introduced a warning in the compiler ([52 fragile constant pattern](http://caml.inria.fr/pub/docs/manual-ocaml/comp.html#sec285)) to discourage people from matching on exception strings to precisely allow this kind of changes, so I think it is ok to change them -- for those whose string payload are not specified in the `.mli`, indeed.

src/batBitSet.ml
```diff
@@ -48,7 +48,7 @@ let capacity t = (Bytes.length !t) * 8
 let empty () = ref (Bytes.create 0)
 
 let create_ sfun c n = (* n is in bits *)
-  if n < 0 then invalid_arg ("BitSet."^sfun^": negative size");
+  if n < 0 then invalid_arg ("BitSet." ^ sfun ^ ": negative size");
```
gasche: While we are at it, I think that `Printf.ksprintf invalid_arg "Bitset.%s: negative size" sfun` would be nicer (and possibly more efficient) here.

src/batFile.ml
```diff
@@ -45,8 +45,8 @@ let perm l =
     ~f:(fun acc x -> acc lor x)
 
 let unix_perm i =
-  if 0<= i && i <= 511 then i
-  else raise (Invalid_argument (Printf.sprintf "Unix permission %o " i))
+  if 0 <= i && i <= 511 then i
+  else invalid_arg (Printf.sprintf "File.unix_perm: Unix permission %o" i)
```
gasche: `Printf.ksprintf invalid_arg` would be nice.

src/batHeap.ml
```diff
@@ -107,35 +107,41 @@ let find_min bh = match bh.mind with
...
+  in
+  let rec find_min_tree ts k =
+    match ts with
+      | [] ->
+          invalid_arg "del_min"
+      | [t] ->
+          k t
+      | t :: ts ->
+          find_min_tree ts (fun u ->
+            if Pervasives.compare t.root u.root <= 0 then k t else k u)
```
gasche: I'm not convinced by this change. I would rather have `find_mind_tree` remain a generic operation (it could be used by other functions later, or even expose to the user without the continuation-taking interface), but take two continuations, `k : 'a t -> 'b` and `kfail : unit -> 'b`, instead of one -- this is standard in functional encoding of search procedures. Then `del_min_tree` could use `(fun () -> raise "del_min")` as the failure continuation.

src/batBitSet.ml
murmour: How about including an actually received `n`? Over years, I've developed a practice that involves stuffing as much information into Invalid_argument messages, because it's an investment that pays off easily. Instead of "Invalid index", my functions throw "Invalid index (expected n >= 0 && n < %d, but got %d)". As for performance, no sane code should ever catch Invalid_argument for normal control flow, as it's akin to a dynamic assertion. I even think of opening an issue about reworking whole Batteries in this style.

UnixJunkie:
I completely agree that error messages should be self explanatory. Up to the point where an end user can understand and fix a problem just by reading the error message.
So, asserts are bad in fact. Much more useful to the end user is a failwith with an as informative as possible error message.

src/batBitSet.ml
murmour: > As for performance, no sane code should ever use Invalid_argument for normal control flow, as it's akin to a dynamic assertion.

Typo: "as it's akin to _suppressing_ a dynamic assertion". I talk about a scenario when the user prefers to catch Invalid_argument instead of ensuring that no bad argument ever gets into the function. My position is that Invalid_argument is about as bad as Assert_failure -- one should never catch it explicitly.

gasche: I agree with indicating the out-of-bound index in the error message. I thought about it during my review and decided not to add more work, but if you are voluntary...

murmour:
Last changes:
* Utility functions used inside `Heap.del_min` are modified according to @gasche's suggestions.
* "raise Invalid_argument" is changed into "invalid_arg" in a couple more places.

As for extending the messages with additional info, I will open a separate issue for this, as it would be better if we first agreed on conventions, extending the style guide.

src/batHeap.ml
```diff
@@ -107,34 +107,40 @@ let find_min bh = match bh.mind with
 *)
 
 
-let rec find_min_tree ts k = match ts with
-  | [] -> failwith "find_min_tree"
+let rec find_min_tree fail ts k = match ts with
```
gasche: I'm sorry for the continued nitpicking, but the two continuation arguments should really follow each other, currently the ordering makes no sense.

In general I would maybe expect the success continuation first and then the failure continuation, but given the current calling style (which is not unrelated to a monadic let),  think that having the failure continuation first makes sense (that is: makes call sites look nicer than with the other choice). It also corresponds to the ordering `None | Some of 'a`, which is not unnatural.

Also, I think it would be nice to name it `kfail` so that people recognize the two-continuations pattern.

murmour: > I'm sorry for the continued nitpicking, but the two continuation arguments should really follow each other, currently the ordering makes no sense.

Whoops: I was too rushed, and didn't notice that the style is clumsy. Please, see my last changes; and thanks for the nitpicking -- details matter.

gasche:
This is nice, thanks!

PR #851, opened 2018-03-20
BatInt.Safe_int.mul performance improvements

murmour:
The previous overflow test from [CERT C](https://wiki.sei.cmu.edu/confluence/display/c/INT32-C.+Ensure+that+operations+on+signed+integers+do+not+result+in+overflow) is notable for avoiding an overflow while checking for overflow; hovewer, since overflows are perfectly defined in OCaml, a shorter test from Hacker's Delight is a better option.

This is a port of my changes from ocaml/ocaml#1520.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,7 +7,7 @@ Changelog
   #799
   (Francois Berenger)
 - Int: optimized implementation of Safe_int.mul
-  #808
+  #808, #851
   (Max Mouratov)
 
 ## v2.8.0 (minor release)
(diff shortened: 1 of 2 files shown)


murmour:
Ping. The formula was validated in ocaml/ocaml#1520. The only difference here is the fast-path check (which is unchanged).

gasche:
Sorry for the delay; I guess that none of the frequent contributors/maintainers here are familiar with this stuff..

PR #852, opened 2018-03-22
List.unique_cmp tweak

murmour:
Self-explanatory.


gasche:
I'm curious now. Does it actually go any faster, or at least use less memory?

murmour:
A set uses one word less memory per the node of a tree, which potentially affects performance. I didn't actually benchmark it.

murmour:
By the way, a hash table should be a much better option here, but that change would require a proper benchmark.

PR #854, opened 2018-03-23
Fix docstrings of List.append and List.concat

murmour:
Self-descriptive.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -55,6 +55,10 @@ Changelog
 - Optimization of List.unique_cmp (using Set instead of Map).
   #852
   (Max Mouratov)
+- Documentation of List.append and List.concat should not include invalid
+  estimates of stack usage.
+  #854
+  (Max Mouratov)
 
 ## v2.8.0 (minor release)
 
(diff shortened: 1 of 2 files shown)


gasche:
I thought that the complexity annotations where about time, not stack space (but then the one on `concat/flatten` seems to be stack space), and am unsure whether they should be removed or clarified/fixed.

murmour:
> I thought that the complexity annotations where about time, not stack space

I assumed that "O(...)" in "Tail-recursive O(length of the first argument)." is about space, because it is placed in the same sentence as "Tail-recursive", and because the non-tail-recursive version of `append` takes exactly as much stack space; but you're right that this also the time complexity! I'll add back the time complexity estimate with a better wording.

As for `concat`, the complexity comment is a [leftover from the OCaml standard library](https://github.com/ocaml/ocaml/blob/b2a91a609090bbb176ed3070d51d99716affda6f/stdlib/list.mli#L96).

murmour:
> I'll add back the time complexity estimate with a better wording.

Done.

PR #855, opened 2018-03-25
Remove mention of IO.flush_out that does not exist.

rixed:


gasche:
Well spotted, thanks!

PR #857, opened 2018-04-02
Fix erroneous mentions of [Different_list_size] in List.mli

### LINKED ISSUE #744: Inconsistency between interface documentation and implementation in BatList
BatList mentions `Different_list_size` at several places (e.g. combine) but actually yields `Invalid_argument`.

murmour:
Self-descriptive. Fixes #744.


gasche:
Thanks!

PR #859, opened 2018-05-01
better ocamldoc for BatHashtbl.find_default

UnixJunkie:
so much better with an example


src/batHashtbl.mli
```diff
@@ -146,8 +146,8 @@ val find_all : ('a, 'b) t -> 'a -> 'b list
     bindings, in reverse order of introduction in the table. *)
 
 val find_default : ('a,'b) t -> 'a -> 'b -> 'b
-(** Find a binding for the key, and return a default
-    value if not found *)
+(** [Hashtbl.find_default tbl key def] find a binding for [key],
```
gasche: finds a binding ... or return [default]

PR #861, opened 2018-05-05
Fix build on OCaml < 4.05

murmour:
This fixes a build issue introduced in #856.


murmour:
(also, this is one of those cases where explicit exception handling looks nice)

gasche:
Good catch, thanks!

thizanne:
Another way to fix it would have been to move `cut_on_char` after the conditional redefinition of `index_from_opt` for `< 4.05`. No strong opinion here though, and thanks for the catch (also, sorry for not thinking to test before 4.05).

PR #862, opened 2018-05-07
Hotfix: a regression in BatInt.Safe_int.mul

murmour:
#851 introduced a regression in one of the tests. I must have forgotten to run the testsuite (blushes).

While copying the formula from the OCaml compiler, I missed out the clause that was added in  ocaml/ocaml#1520, because I forgot to switch branches.

Here is the formula from OCaml, for reference:
```ocaml
let no_overflow_mul a b =
  not ((a = min_int && b < 0) || (b <> 0 && (a * b) / b <> a))
```

### DIFF
--- a/src/batInt.ml
+++ b/src/batInt.ml
@@ -258,7 +258,8 @@ module BaseSafeInt = struct
   let mul (a: int) (b: int) : int =
     let open Pervasives in
     let c = a * b in
-    if (a lor b) asr mul_shift_bits = 0 || b = 0 || c / b = a then
+    if (a lor b) asr mul_shift_bits = 0
+    || not ((a = min_int && b < 0) || (b <> 0 && c / b <> a)) then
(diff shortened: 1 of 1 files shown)


gasche:
I am very happy that you are proving my point on unit tests in an unexpected way :-)

(We used to have Continous Integration running on master thanks to @c-cube, the service might have been disabled. We should try to setup some Travis CI...)

c-cube:
Oh, sorry about the CI. My jenkins broke one time too many for my patience, and I use travis for a few of my own projects now :disappointed:

murmour:
> I am very happy that you are proving my point on unit tests in an unexpected way :-)

Point taken :-). Another good point is that I should have audited the change more carefully.

murmour:
To redeem myself, I'll include a formal proof with with my next PR.

PR #863, opened 2018-05-14
added Unix.time_it

UnixJunkie:
there is no unit test, I can add one if people are interested.
Along the lines: BatUnix.time_it (fun () -> Unix.sleep 1) ~= 1

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,9 @@ Changelog
 
 ## NEXT_RELEASE
 
+- add Unix.time_it: ?n:int -> (unit -> 'a) -> float * 'a
+  #863
+  (Francois Berenger, review by Gabriel Scherer)
 - add `BatString.cut_on_char : char -> int -> string -> string`
   (Kahina Fekir, Thibault Suzanne, request by François Bérenger)
   #807, #856
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,7 +3,7 @@ Changelog
 
 ## NEXT_RELEASE
 
-- add Unix.time_it: ?n:int -> (unit -> 'a) -> float * 'a
+- add Unix.time_it: (unit -> 'a) -> float * 'a
   #863
   (Francois Berenger, review by Gabriel Scherer)
 - add `BatString.cut_on_char : char -> int -> string -> string`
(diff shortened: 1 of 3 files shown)


gasche:
Yes, a test would be nice.

I also wonder about
- the name, maybe `with_time` would be a bit better?
- the interface, do we want `(unit -> 'a) -> float * 'a`, or `('a -> 'b) -> 'a -> float * 'b`? (I tend to prefer your proposal, `(unit -> 'a) -> float * 'a`.)

gasche:
But then which time do you return, the total or divided by `n`?

UnixJunkie:
About the name: exec_time would also be another possible name.

UnixJunkie:
the total time of course; we are not in a Stats module. :wink:

UnixJunkie:
I added a unit test. I hope other maintainers will forgive me for making 'make test' take at least 0.01s more from now on.

UnixJunkie:
OK, I added the optional ~n parameter, defaulting to 1.
This would be my final proposal, changelog update included.

UnixJunkie:
it would also be possible to return an array of floats instead of just one, if people really want to do stats after.
That would be a little bit a lower level interface though.

gasche:
I don't really see the point of the `~n:int` parameter:

- this introduces several uncertainties in behavior, such as "which result is returned?" and "is the total time or the time of just one iteration returned?"
- people can always write `for i = 1 to n do ignore f () done` by hand if they want
- I don't see a clear use-case for it

There is the problem, with or without the parameter, that the precision of this function is going to be awful for short computations. This should be pointed out explicitly in the documentation. But if we want to do any better than "only use it when precision doesn't matter that much", I think what you want is an easy-to-use benchmarking library (that can also comment on average, variance etc.), rather than a timing functions with *some* extra features.

UnixJunkie:
Ok, I removed the ~n.

UnixJunkie:
So, good? Not good?

UnixJunkie:
@c-cube any opinion?

c-cube:
I'm not sure what the goal is, given the poor precision of gettimeofday. Maybe this would be more suited to oclock or ptime?

UnixJunkie:
I personally run computational experiments. They take long time usually (like several seconds, sometimes minutes or hours).
The precision of gettimeofday is far enough for me.

UnixJunkie:
Closing this stalled PR of mine which did not raise a lot of enthusiasm from other maintainers.

PR #865, opened 2018-05-30
Map.equal used (=) instead of argument

vogler:
Map.equal does not use its argument for checking equality:

~~~ocaml
# BatMap.(equal (fun x y -> print_endline "eq"; x=y) (add 1 2 empty) (add 1 2 empty))
- : bool = true
~~~

The functorized versions are fine.


gasche:
Ouch, thanks!

PR #873, opened 2018-08-23
Current module name seems Extlib, not Extlibcompat

catalin-hritcu:
https://github.com/ocaml-batteries-team/batteries-included/blob/master/src/extlib.ml


UnixJunkie:
True, thanks!

PR #874, opened 2018-08-30
Batoption equality

gregory-nisbet:
Hi everyone,

I added an implementation for `BatOption.equal`, based on the function in `BatSeq` which seems to be the most general API for a function called `equal` in the project. 

I also added a test in the non-property-based portion in the test suite.

### DIFF
--- a/src/batOption.ml
+++ b/src/batOption.ml
@@ -75,6 +75,12 @@ let default_delayed l = function
   default_delayed (fun () -> assert false) (Some 4) = 4
 *)
 
+let equal ?eq:(eq0=Pervasives.(=)) x y =
+  match (x, y) with
+  | None, None -> true
+  | Some x, Some y -> eq0 x y
+  | _ -> false
+
 let is_some = function
   | None -> false
   | _ -> true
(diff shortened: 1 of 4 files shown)


UnixJunkie:
Looks useful, thanks.
Can you inline the tests in batOption.ml rather than adding a new file.
The unit test should be about four lines this way.

gregory-nisbet:
@UnixJunkie 

On further inspection, `equal` is an exact duplicate of the function `BatOptiol.eq` which already exists.

I'm not sure what I should do. I got the task from the list of low hanging fruit for new people.

UnixJunkie:
Ok, let's close this PR then.
Sorry.
Where did you get the page of low hanging fruits?
The bug tracker
https://github.com/ocaml-batteries-team/batteries-included/issues
has many things.
Maybe it is more up to date and you can pick something from it.

gregory-nisbet:
Ici:

https://github.com/ocaml-batteries-team/batteries-included/wiki/%22Junior-tasks%22-for-batteries

It might still be useful to create an `equal` alias for `eq` to bring it in line with the other modules.

Does Batteries have general-purpose interfaces for `equal-able`, `comparable`, &c things?

Also, thank you for the link to the bug tracker.

UnixJunkie:
If many other modules have an equal, then an alias looks like a good idea.
But other maintainers might voice differently.

PR #876, opened 2018-09-08
release 2.9.0 with support for OCaml 4.07.0

gasche:
This release candidate is unusual in that it does *not* pass the "test-compat" test under 4.07.0, which checks that all the functions of the standard library modules are exported by Batteries. For now, I have only implemented the minimal changes to compile and work correctly under 4.07.0.

Given that my own contribution time is limited these days, and that the standard library is moving faster than before, my plan is to stop backporting new functions under older batteries release -- the new project [stdcompat](https://github.com/thierry-martinez/stdcompat) by @thierry-martinez is taking care of that in a better-defined and lighter way than Batteries was.

Of course, others are warmly welcome to contribute support for new stdlib functions in Batteries modules.


UnixJunkie:
I remove what I said previously, because stdcompat has non tail rec code.

gasche:
Are you allright with the idea of releasing 2.9.0 without Seq support (or support for other 4.07 stdlib additions)? We can still release 2.10.0 with more complete support if we wish.

Personally I would rather go for an incomplete but 4.07-compatible release than not have a 4.07-release at all. We should also discuss whether we plan to drop the completeness requirement in the long term (personally I think that could be more reasonable), or see this as a one-time slip.

UnixJunkie:
I agree to ship; better something now than more never.
This can be locking for users who want to use 4.07.0.
Also, having releases from time to time keeps the project alive.
In general, I am OK to decrease the code size, to relieve the maintenance burden.
We have few maintainers and contributors, so...

thierry-martinez:
@UnixJunkie I can understand that it could be better for batteries to avoid a dependency on stdcompat, but I don't see the point of worrying about non tail rec code: OCaml standard library has non tail rec code as well, and batteries ”depend“ on it necessarily. The point of using stdcompat would be to ensure that newer standard library symbols are available (like Seq.*), even on old versions of OCaml.

UnixJunkie:
I had a look at the diff of this PR.
Apparently, this is only boiler-plate changes, so I am OK.
Also, I confirm that make test works in 4.07.0 with this branch.

UnixJunkie:
@thierry-martinez I am afraid of making batteries depend on something that exposes
non tail recursive functions and that users may think are provided by batteries.
But maybe I am worried for no good reason and this could remove us many 
lines of pre-processing boilerplate code.

gasche:
Thanks @UnixJunkie for your review. I will merge and send the release on opam, but please let's feel free to discuss these issues here.

(Personally I am not opposed to adding stdcompat as a dependency as long as it doesn't itself drag many dependencies. What I think is important is that proper tests are written for the functions we add, either in stdcompat or Batteries, because the stdlib isn't currently doing a good enough job at that. @thierry-martinez found the last stdlib bug, so I guess stdcompat does have tests :-)

UnixJunkie:
stdcompat has optional dependencies:
depopts: ["result" "seq" "uchar"]

PR #885, opened 2018-12-09
Fix Dllist.skip

### LINKED ISSUE #884: BatDllist.skip not according to doc/spec
The doc for BatDllist.skip indicates that negative counts will traverse over the previous node:

> val skip : 'a node_t -> int -> 'a node_t
> skip n i Return the node that is i nodes after node n in the list. If i is negative then return the > node that is i nodes before node n in the list. This is an O(N) operation.

Whereas the implementation always traverses the next element in the list.

Firobe:
Reporting fix from Extlib : skip had the same behavior no matter the sign of the parameter  
Closes #884

### DIFF
--- a/src/batDllist.ml
+++ b/src/batDllist.ml
@@ -173,14 +173,14 @@ let next node = node.next
 let prev node = node.prev
 
 let skip node idx =
-  let m = if idx > 0 then -1 else 1 in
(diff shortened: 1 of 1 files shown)


gasche:
Thanks for the fix!

Issue #888, opened 2019-03-20
BatList.fold_while


UnixJunkie:
I think we miss this function.
It should do the fold as long as the given predicate holds, and return the unconsumed
part of the list once it does not anymore.

UnixJunkie:
This is not the first time that I need this.
There is one in containers, for example.
I think there is in haskell too IIRC.
If other maintainers are interested, I can send some code.

gasche:
Sure, why not. What's the function signature (can you link to the Contains version)?

UnixJunkie:
The signature in containers (from here: https://c-cube.github.io/ocaml-containers/last/containers/CCList/index.html):
```
val fold_while : ('a -> 'b -> 'a * [ `Stop | `Continue ]) -> 'a -> 'b t -> 'a
```
I just provide it for info.
I don't think it is explicit enough.

UnixJunkie:
My proposal is here:
https://github.com/ocaml-batteries-team/batteries-included/pull/889

UnixJunkie:
PR was merged

PR #889, opened 2019-03-22
added BatList.fold_while

UnixJunkie:


thizanne:
Without introducing an additional mutable state, this signature doesn't
allow you to, say, "sum the elements of the list until the sum is greater
than 42" : the predicate can only check the list element value. I'd say
having the more expressive version is valuable.

UnixJunkie:
@thizanne what is the signature that you are thinking about?

thizanne:
You have several ways of doing it, and quoted one in #888 : 

```val fold_while : ('a -> 'b -> 'a * [ `Stop | `Continue ]) -> 'a -> 'b t -> 'a```

I'm not sure why you find it "not explicit enough". The important part is that the function that decides when the fold has to be stopped has access to the value of the accumulator.

Iterating from your proposal:
```val fold_while : ('a -> bool) -> ('acc -> 'a -> 'acc) -> 'acc -> 'a list -> 'acc * 'a list```

We can add this information (the value of the accumulator) to the predicate:
```val fold_while : ('acc -> 'a -> bool) -> ('acc -> 'a -> 'acc) -> 'acc -> 'a list -> 'acc * 'a list```

Then instead of two functions that take the same parameters, we require one function that will return a tuple. This has several advantages: it makes it explicit in the type that these operations (the construction of the accumulator and the predicate testing) are being done on the same operands, and since they may do the same computation, it's more convenient to define them as one function rather than to artificially split it:
```val fold_while : ('acc -> 'a -> bool * 'acc) -> 'acc -> 'a list -> 'acc * 'a list```

And you get the function of Containers except it uses `bool` instead of ```[`Stop | `Continue]``` (which is a design choice, I personally like the latter more but we should be consistent with other functions of that kind in batteries -- I didn't check).

UnixJunkie:
Ok, got it.
I will update the PR.

UnixJunkie:
I updated the PR.

UnixJunkie:
Anybody to review and merge this?

PR #892, opened 2019-04-19
Hashtbl merge

rixed:


src/batHashtbl.mlv
```diff
@@ -174,10 +174,15 @@ let map_inplace f h =
   in
   BatArray.modify loop (h_conv h).data
 
-(*$= map_inplace & ~printer:(IO.to_string (List.print Int.print))
+(* Helper functions to test hashtables which values are integers: *)
+(*$inject
+  let printer = IO.to_string (List.print Int.print)
+  let to_comparable h = values h |> List.of_enum |> List.sort Int.compare
```
gasche: `to_sorted_list`?

src/batHashtbl.mlv
```diff
@@ -357,10 +362,108 @@ let filter_map_inplace f h =
...
+let merge f a b =
+  let res = create (max (length a) (length b)) in
+  let may_add_res k v1 v2 =
+    BatOption.may (add res k) (f k v1 v2) in
+  iter (fun k v1 ->
+    match find b k with
+    | exception Not_found ->
+        may_add_res k (Some v1) None
+    | v2 ->
+        may_add_res k (Some v1) (Some v2)
```
gasche: cannot we use `may_add_res k (Some v1) (find_opt h2)`?

src/batHashtbl.mlv
```diff
@@ -357,10 +362,108 @@ let filter_map_inplace f h =
...
-(*$= filter_map_inplace & ~printer:(IO.to_string (List.print Int.print))
+(*$= filter_map_inplace & ~printer
   (let h = Enum.combine (1 -- 5, 1 -- 5) |> of_enum in \
    filter_map_inplace (fun _ x -> if x>3 then Some (x+1) else None) h ; \
-   values h |> List.of_enum |> List.sort Int.compare) [5; 6]
+   to_comparable h) [5; 6]
+*)
+
+
+let merge f a b =
```
gasche: This should be `h1`, `h2` by analogy with `v1`, `v2`.

src/batHashtbl.mlv
```diff
@@ -357,10 +362,108 @@ let filter_map_inplace f h =
...
+     merge (fun _ _ r -> r) a b |> to_comparable)
+  [1; 2; 3] \
+    (let a = Enum.combine (1 -- 3, 1 -- 3) |> of_enum and b = Enum.combine (3 -- 6, 13 -- 15) |> of_enum in \
+     merge (fun _ l _ -> l) a b |> to_comparable)
+  [13; 14; 15] \
+    (let a = Enum.combine (1 -- 3, 1 -- 3) |> of_enum and b = Enum.combine (3 -- 5, 13 -- 15) |> of_enum in \
+     merge (fun _ _ r -> r) a b |> to_comparable)
+  [] \
+    (let a = Enum.combine (1 -- 3, 1 -- 3) |> of_enum and b = Enum.combine (3 -- 5, 3 -- 5) |> of_enum in \
+     merge (fun _ _ _ -> None) a b |> to_comparable)
```
gasche: `merge` can express both `inter` and `union`, and it would be nice to see those appear in tests. We could consider having qcheck-style property tests (`equal (inter a a) a`, `equal (union a a) a`, `equal (union a empty) a`, `equal (inter a empty) empty`, `equal (inter a b) (inter b a)`, `equal (union a b) (union b a)`, etc.

src/batHashtbl.mlv
rixed: Indeed!

src/batHashtbl.mlv
rixed: I gave it a try but I'm not very satisfied with it. Please have a look.
The problem is that it gives a misleading feeling of confidence, but actually the union and intersection of hash tables when multiple bindings to the same keys is allowed are not as natural as it seems. Indeed, what's the intersection of `{1->2}` and `{1->3}` ? Is it `{1->2}`, `{1->3}`, `{1->[2;3]}` or `{}` ? And what's the union of `{1->1}` with `{1->2}` ? Is it `{1->[1; 2]}` ? but then, what's the union of `{1->1}` and `{1->1}` ? is it `{1->[1;1]}` ?

Implementing those tests actually made me realise that `Hashtbl.iter` and `Hashtbl.keys` were not working as I've always assumed they would. For instance, I've always though `iter` would send multiple bindings to the same key successively, but it does not (it iterate over the bucket but keys might not be next to each others in the bucket), and I've also always though keys would return the keys only once but it actually return the keys several times in case of multiple bindings. So, quite useful  :)

At the end of the day (literally), I updated the documentation of [merge] to guard against using it in case of multiple bindings.

WDYT?

UnixJunkie:
The doc for merge_all should specify the order of the values in the two lists.
The ChangeLog file should be updated also, if this is your final proposal.

rixed:
> The doc for merge_all should specify the order of the values in the two lists.

Yes you are right, and actually I think the proposed `merge_all` was not doing the right thing. It's now adding the entries back to front so that the head of the returned list become the "most recent" binding, as returned by `find_all`.

Also updated the ChangeLog.

rixed:
So what's the status regarding this?

UnixJunkie:
I still have some questions/remarks:

Is it really necessary to have two times the code for merge and merge_all repeated in the file?

Also, the unit tests are pretty long (longer than the two functions introduced).
Can't they be made much shorter and readable?
For example, it seems that the same hashtables are declared several times.
Maybe factoring those declarations would help making the tests shorter and hence more readable.

rixed:
> Is it really necessary to have two times the code for merge and merge_all repeated in the file?

For some value of "necessary" yes. The hash types are not the same. If you look at other functions you will see that they are all repeated (when not, that's because they are missing from one of the alternatives).

> Also, the unit tests are pretty long (longer than the two functions introduced).
> Can't they be made much shorter and readable?
> For example, it seems that the same hashtables are declared several times.
> Maybe factoring those declarations would help making the tests shorter and hence more readable.

Factored out a few hashes and a function that were reused several times for readability.
It doesn't change much, though. The policy governing tests eludes me. Shall I delete (some of) the tests? If anything, I think in general inline tests are too short: they lack a one line comment explaining, for each test, what's tested.

UnixJunkie:
Thanks. I hope I did not sound like the annoying guy. I was just trying to provide some constructive feedback on the PR.

rixed:
> Thanks. I hope I did not sound like the annoying guy. I was just trying to provide some constructive feedback on the PR.

No worries! Esp. the remarks were totally valid. Not answering would have been annoying :)

PR #896, opened 2019-05-14
added BatRandom.full_self_init

UnixJunkie:
For numerical experiments, I need to be able to initialize the PRNG with more entropy.

### DIFF
--- a/src/batRandom.ml
+++ b/src/batRandom.ml
@@ -33,6 +33,23 @@ let float     = Random.float
 let bool      = Random.bool
 let char ()   = Char.chr (int 256)
 
+let with_in_file fn f =
+  let input = open_in_bin fn in
+  let res = f input in
+  close_in input;
+  res
+
+let full_self_init n =
+  if n <= 0 then invalid_arg "BatRandom.full_self_init";
+  let prng_fn = "/dev/urandom" in
+  if not (Sys.file_exists prng_fn) then
+    failwith ("BatRandom.full_self_init: no such file: " ^ prng_fn);
+  let rand_ints =
+    with_in_file prng_fn (fun input ->
+        Array.init n (fun _ -> input_byte input)
+      ) in
+  full_init rand_ints
+
 let full_range_int =
   if Sys.word_size = 32 then (* need 31-bits of entropy, bits() gives 30 *)
     fun () -> if bool () then - (bits ())-1 else bits ()
--- a/src/batRandom.mli
+++ b/src/batRandom.mli
@@ -46,6 +46,13 @@ val self_init : unit -> unit
 (** Initialize the generator with a more-or-less random seed chosen
     in a system-dependent way. *)
 
+val full_self_init : int -> unit
+(** [full_self_init n] initialize the generator with [full_init]
+    using [n] bytes read from /dev/urandom; each individual byte being
+    converted to an integer.
+    @raise Invalid_arg if [n] <= 0.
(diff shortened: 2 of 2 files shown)


UnixJunkie:
If someone knows how to make this work on windows, in pure OCaml, contribution welcome.

gasche:
`Random.self_init` from the standard library, which we reuse, already uses `/dev/urandom` when available. What is your reason for using this new function instead?

UnixJunkie:
to easily ask for a longer random_seed.
I will have a look at "caml_sys_random_seed" though.

UnixJunkie:
caml_sys_random_seed reads only 12 bytes from /dev/urandom.
While, I want the user to be able to control how many bytes are read from it.
E.g. I will personally use this new function with at least 20 or so as the int parameter.

gasche:
My guess as to why it only reads 96 bits would be that this is the size of the state used by the OCaml PNRG, so adding any more bits is useless. But I have not checked.

gasche:
So I looked and: the code for `Random.full_init` is rather obscure (and not clearly trying to maximize the entropy of the initialization), but it seems that it could use up to 55 random 31-bit integers worth of entropy for maximal initialization, which is more than what is currently read. But the entropy-mixing logic is weird and I don't think one can easily reason about how much entropy one actually gets by passing anything longer than that. I would prefer to clarify that with the upstream instead of providing another function to users that nobody knows how to use correctly (or whether it's possible to use it correctly).

UnixJunkie:
For numerical simulations, it is nice to be able to initialize the PRNG with a lot of bits, so that
two distinct simulations have a low chance of being done with the same seed.

UnixJunkie:
I think, a fair critic of my proposed contribution would be that I create 0 to 255 integers, instead of properly creating 31 or 63 bit OCaml integers (depending on the architecture) using the pseudo random bytes that were read from /dev/urandom.
"instead of providing another function to users that nobody knows how to use correctly" is not very constructive/fair.
Here is a usage example:
```
BatRandom.full_self_init 20
```
And here is the core logic:
```  let rand_ints =
    with_in_file prng_fn (fun input ->
        Array.init n (fun _ -> input_byte input)
      ) in
  full_init rand_ints
```
That's fairly easy to understand (compared to the weird cooking done in "caml_sys_random_seed").
And, it just assumes that Random.full_init works.

gasche:
My point is that full_init works in complex ways; it does *not* seem to assume that its input is truly random. It performs iterated MD5 hashes on the string representation of integers it is passed, and you probably *cannot* assume that sending 192 bits of entropy instead of 96 of entropy gives you twice the entropy on the random generator state. It's not a criticism of your code, but your documentation would have to be "if you see repetitions, then maybe try passing a larger `n`, no guarantees on the amount of entropy it provides", and that is not very helpful.

Have you encountered a concrete issue with the amount of entropy provided by `Random.self_init`, and have you observed improvements with the function you propose?

UnixJunkie:
I have seen simluations which were obviously using the same seed; I was using Random.self_init ().
I feel Random.full_init is not "batteries included" enough.
So, I propose full_self_init, to have something between self_init and full_init.
Better than self_init in terms of the amount of entropy, and better than full_init in terms of end-user experience (he doesn't have anymore to look at how to get random integers).

Another fair critic would be that my proposition doesn't work under MS windows; and I have not the slighest idea on how to fix that one.

UnixJunkie:
I am closing this stalled PR of mine.

PR #898, opened 2019-07-09
[RFD] BatTuple.Tuple?.split [do not merge]

jrochel:
NOTE: This PR does not compile!
 
The fact that these functions do not exist somewhere is clearly an annoyance. Now, I wanted to inquire whether this is the right place to put these functions or whether this is an inadmissible commingling of concerns (list stuff in BatTuple).

If not, I'd supply a PR with a working implementation.

### DIFF
--- a/src/batTuple.ml
+++ b/src/batTuple.ml
@@ -70,6 +70,8 @@ module Tuple2 = struct
     let comp = cmp1 a c in
     if comp <> 0 then comp else cmp2 b d
 
+  let split = BatList.split
+
   open BatOrd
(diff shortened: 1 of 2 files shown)


rixed:
Shouldn't the split functions use enums rather than lists?

jrochel:
Right. I guess you're right. But the core question remains...

rixed:
I have no opinion regarding the core question. I never needed those functions so I never noticed they were missing; and I can't say whether I would have expected to find them there. I seldom use the TupleN modules but for printing, anyway.

PR #899, opened 2019-07-11
4.08 compatibility

gasche:


UnixJunkie:
We have CI tests for batteries?!

gasche:
It looks like we don't right now, but the Travis files are there so I'm maintaining them when it is easy to do so. (Would you like to work on re-enabling them?)

UnixJunkie:
I don't have time right now.
But, I will create an issue to track it.

gasche:
I went ahead and merge the PR -- it's pretty safe from a backward-compatibility point of view.

We should try to do a release soon-ish to get 4.08 support in opam-repository.

PR #901, opened 2019-07-16
Corrected BatDeque.is_empty documentation.

zepalmer:
`BatDeque.is_empty`'s documentation states that it returns the opposite of what it actually returns.  This pull request corrects the documentation.  The behavior of the function `is_empty` is correct: `BatDeque.is_empty BatDeque.empty` ⇒ `true` (OCaml 4.07.1, Batteries 2.9.0).


gasche:
Indeed! Thanks.

Issue #908, opened 2019-09-26
BER MetaOCaml cannot compile batteries 


rixed:
Because of the `>.` operator from BatSet.Infix.
I'd like to rename '<.' and '>.' into '<<' and '>>' for instance, as a breaking change in the next major release.
WDYT?

UnixJunkie:
There is a slight problem: someone should really invest significant manpower in the next _major_ release, which is very long overdue.
I am OK with your proposition, though.

rixed:
Is there a list of things to be done for the next release written down somewhere, or are we speaking about [those bugs tagged for next release](https://github.com/ocaml-batteries-team/batteries-included/issues?q=is%3Aissue+is%3Aopen+label%3A%22for+next+major+release%22) ?

UnixJunkie:
When I was triaging our issues list, yes I tagged them as you found out.

UnixJunkie:
We also have PRs with this kind of tag:
https://github.com/ocaml-batteries-team/batteries-included/pulls?q=is%3Aopen+is%3Apr+label%3A%22for+next+major+release%22

rixed:
Good to know that the bug tracker is reliable in this regard. Will start to work my way through the list, as many seems easy enough.

rixed:
Fixed by 9d3ab53cc06f2aee0920c6b350b3c297d00f0370

PR #909, opened 2019-09-26
Fix compilation by BER MetaOCaml

rixed:
Set.Infix used to have operators `<.` and `>.` (for strict
subset and strict superset), but `>.` is also used as a code
block end-quote by BER.

Therefore the infix operators have been renamed into
`<<` and `>>`, which breaks the homogeneity of those operators somewhat,
but:

- We'd better not rename all those infix operators;
- Those infix operators are going to be ugly no matter what.

Ref #908


rixed:
I just realised that '<<' and '>>' are quite likely to be used as quoting operators by some other tool.
should we just rename `>.` into `.>` maybe?

rixed:
Indeed, it didn't take long to find one: https://oliviernicole.github.io/about_macros.html

gasche:
I agree it's an important question and I don't know what is a good fix. I would consider the following:

- removing the infix operators for Set altogether
- or using a different suffix than `.`

UnixJunkie:
I like the removing the operators altogether option.

rixed:
Going with that suggestion then.

UnixJunkie:
Sometimes, I think batteries is too big/fat.
It is nice if we find ways to make it smaller/simpler.

rixed:
> Sometimes, I think batteries is too big/fat.

@UnixJunkie I think there is a discussion to have about how to improve stdlib now that the stdlib itself is improving. Larger and larger part of Batteries will be copy-pasted from/to the stdlib.
Maybe we should make batteries a thinner overlay on top of stdlib without trying to bring back newer function to user using older compiler?

UnixJunkie:
We don't have much manpower usually, so not "trying to bring back newer functions to users using older compiler" is very likely (and I think it already happened recently).
However, if the new stdlib functions are not tail-rec, we should still re-implement them in batteries.

PR #910, opened 2019-09-27
Add BatFilename

### LINKED ISSUE #445: no BatFilename in batteries?

rixed:
Also with `Filename.split_extension`, provided I've understood the
requirement for this function.

Closes #445

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ src/batPrintf.mli
 src/batFormat.mli
 src/batSys.mli
 src/batBigarray.mli
+src/batFilename.mli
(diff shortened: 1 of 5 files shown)


UnixJunkie:
thanks for this contribution.

UnixJunkie:
The compilation is broken by this addition under 4.02.3.
What is the older version we claim to support?

UnixJunkie:
this is not trivial to fix: some more preprocessing directives are needed in batFilename.mliv

UnixJunkie:
I don't have time today...

Issue #839, opened 2018-02-22
Result.result type incompatible with ocaml Pervasives.result 


sapristi:
Hello, 
I noticed the Batteries result type:
```
 type ('a, 'b) BatPervasives.result =
  | Ok  of 'a
| Bad of 'b
```
is incompatible with OCaml Pervasives.result type :
```
 type ('a, 'b) Pervasives.result =
  | Ok  of 'a
 | Error of 'b
```
Which make it difficult to use a library not using batteries. 
(In fact I stumbled upon an error caused by this months ago and just realised where the problem was).
Is there any reason for this ? Although modifying the entire Batteries to take this into account might be overkill, it would make things more usable.

gasche:
The `result` types in Batteries has existed for years, while the OCaml standard library only included it recently. There is no way supported by the OCaml language to change a constructor name (`Bad` -> `Error`) without breaking user code written against old versions of the library. Batteries tries fairly hard to keep code written in the past working, and it is not clear to me that the advantages of the rename outweigh the compatibility costs.

sapristi:
Ok well so it is then !

c-cube:
@gasche a "simple" solution would be to add a new module for the standard `result` type, deprecate the current one, and remove it in a few versions. I think you might underestimate the cost of being incompatible with literally every other OCaml project (same for IO).

UnixJunkie:
@sapristi if you don't do 'open Batteries', is this still a problem?

thizanne:
I'm for aiming to be compatible with the standard OCaml definition. Maybe we could consider this for next major release™, possibly with a deprecation scheme like @c-cube proposes to not break everything at once ?

rixed:
> I think you might underestimate the cost of being incompatible with literally every other OCaml project

It's not incompatible, it's just force users to not open Betteries.
Actually, if we included the OCaml Result module in BatPervasives.Legacy, that would make it
easier to deal with libraries returning "legacy" Result (quoted here because, as gasche have said, the Result module appeared in OCaml only at v4.08.0).

So I propose to add Legacy.Result as a starting point (this does not have to wait for next release).

Then, for the next release, to rename bateries' Result.Bad into Error, skipping the deprecated period that's just hand-waving in my experience.

UnixJunkie:
I like rixed's bold suggestion.

UnixJunkie:
fixed in v3

PR #913, opened 2019-09-27
Add Legacy.Result for OCaml >= 4.8.0

rixed:
So that users can pattern match Result.t returned from some library using
Legacy.Result.Error while Batteries is opened.

Ref #839


rixed:
This does not have to wait the next major release.

PR #916, opened 2019-09-27
List.sum [] is now 0

### LINKED ISSUE #519: List.sum fails on empty list

rixed:
And similarly for fsum.

Closes #519

### DIFF
--- a/src/batArray.mlv
+++ b/src/batArray.mlv
@@ -592,12 +592,13 @@ let min_max a =
     try ignore (min_max [||]); false with Invalid_argument _ -> true
 *)
 
-let sum = reduce (+)
-let fsum = reduce (+.)
+let sum = fold_left (+) 0
+let fsum = fold_left (+.) 0.
 
 (*$T sum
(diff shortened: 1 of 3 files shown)


UnixJunkie:
I would prefer that '0.0' is used rather than '0.'.
Because, '0.0' is more explicit/easier to read (you don't miss easily that there is a '.' then).

rixed:
If you can't spot the trailing dots then you will have a hard time with floats in OCaml, as every float operators end with a dot. :)
This source file already have plenty of literal `0.` for only a handful of `0.0` so I think it's OK.

UnixJunkie:
Improving the code quality little by little does not hurt.

rixed:
True, but quality is also about consistency. I'll do a global sed later for that.

Issue #845, opened 2018-03-02
String.nsplit is not consistent with String.split_on_char


thizanne:
`String.nsplit` returns the empty list is the string parameter is empty. `split_on_char` never returns an empty list. 

Should we consider having them both behave the same, or is it best left as it is ? This should only be considered for next major release anyway, since it breaks some code.

thizanne:
Actually `split_on_char` seems broken, it does return the empty list when applied to the empty string. See #846

gasche:
`split_on_char` and `nsplit` have incompatible specifications, so unfortunately there is not much we can do to make them "coherent" before v3 -- but I believe that it's a better specification to never return an empty list. We should fix our `split_on_char` implementation now.

rixed:
Not sure I understand this or if this is outdated already, but on batteries master I have:

```
utop # open String;;
utop # let s1 = split_on_char '|';;
val s1 : string -> string list = <fun>
utop # let s2 = nsplit ~by:"|";;
val s2 : string -> string list = <fun>
utop # s1 "" = s2 "";;
- : bool = true
utop # s1 "foo" = s2 "foo";;
- : bool = true
utop # s1 "foo|" = s2 "foo|";;
- : bool = true
utop # s1 "foo|bar" = s2 "foo|bar";;
- : bool = true
utop # s1 "|bar" = s2 "|bar";;
- : bool = true
utop # s1 "|" = s2 "|";;
- : bool = true
```

@thizanne what am I missing?

thizanne:
Short answer in a hurry: #846

rixed:
Ah I see.
So we want both BatString.split_on_char and BatString.nsplit to return `[""]` on an empty string.

rixed:
One of my own program just got broken by this change, and as this change of API does not trigger any warning or anything, it took me an actual debugging session to figure out that the tests broke because of this change, and I'm the one who have implemented it. Also, I'm lucky that the bug has been caught by the tests. But I cannot help but think that my code is not the only one that will be subtly and silently broken.

I think it's better to change the API in a way that breaks compilation rather that runtime behavior.

Should we consider introducing a new function or a new optional parameter or anything like that to at least make sure users are aware of the change after an upgrade?

UnixJunkie:
Maybe revert the change and resolve this issue as "won't fix".

rixed:
We cannot merely `wontfix` this.
Two things are not acceptable, in my opinion:

[...]

PR #924, opened 2019-10-09
BatString: Another take on nsplit changes

rixed:
Recently, [nsplit] behavior wrt. empty strings was changed to return a
unique empty string instead of an empty list, to match both the
specifications and [split_on_char] behavior.

The problem with that is it could silently broke user programs.

Therefore, this patch reverts the change to [nsplit] (but adds a
deprecation warning) and introduce the new [split_on_string] with the
intended behavior.

Note: [split_on_char] has the same issue, but there is no good fix for
it as not behaving like the Stdlib seems as problematic as not behaving
like former Batteries version. [split_on_char] change is thus regarded
as a bugfix as opposed to an API change.

Ref #845

### DIFF
--- a/src/batString.mliv
+++ b/src/batString.mliv
@@ -787,12 +787,23 @@ val rsplit : string -> by:string -> string * string
 *)
 
 val nsplit : string -> by:string -> string list
+##V>=4.2##  [@@ocaml.deprecated "Use split_on_string instead."]
 (** [nsplit s sep] splits the string [s] into a list of strings
     which are separated by [sep] (excluded).
-    [nsplit "" _] returns a single empty string.
-    Note: prior to NEXT_RELEASE [nsplit "" _] used to return an empty list.
+    [nsplit "" _] returns an empty string.
 
     Example: [String.nsplit "abcabcabc" "bc" = ["a"; "a"; "a"; ""]]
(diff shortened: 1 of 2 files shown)


UnixJunkie:
It looks OK, but there is a typo in a comment: split_in_sting

rixed:
Thank you!
By the way, `split_on_string ~by` looks so wrong. Should we rename `~by` into `~on` ?

UnixJunkie:
But, BatString.split also uses ~by ...

UnixJunkie:
BatString.nsplit too uses ~by, by the way.
I don't think renaming those is a good idea.
I guess people can live with those "bad" names.

rixed:
> BatString.nsplit too uses ~by, by the way.

Yes but it does not have "on" in its name, so there can be no conflict here.

> I guess people can live with those "bad" names.

Sure, but since [split_on_string] is a new function.
No strong feeling though, if yo feel it's better not to rename then that's fine with me too.

rixed:
So what's the decision for that one?

UnixJunkie:
There is still a typo (split_on_sting) in a comment.

rixed:
Indeed, fixed.

PR #927, opened 2019-10-23
Build with -strict-sequence

Armael:
This PR enables the `-strict-sequence` flag when building batteries. A few fixes were needed.

- for functions of `BatIO` in an `Incubator` module, I opted to restrict the type in the interface;
- for other functions, I instead inserted calls to `ignore` so as to preserve the library interfaces.

### DIFF
--- a/_tags
+++ b/_tags
@@ -12,4 +12,4 @@ true: package(bytes), warn_-3, bin_annot
 <src/batOpaqueInnerSys.*>: opaque
 true: safe_string
 true: no_alias_deps
-
+true: strict_sequence
--- a/src/batIO.mli
+++ b/src/batIO.mli
@@ -951,7 +951,7 @@ module Incubator : sig
       ?last:string ->
       ?sep:string ->
       ?indent:int ->
-      (Format.formatter -> 'a -> 'b) -> Format.formatter -> 'a array -> unit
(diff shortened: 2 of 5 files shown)


UnixJunkie:
Is this available on all compiler versions?

What are the benefits of using this option?

UnixJunkie:
Thanks for the explanations.
This could be merged then, but it has to be reflected in v3 too.

Armael:
>I am not sure about @rixed 's plan for the v3 branch.

It might be simpler to merge this PR into master and then to merge master into v3 actually?

> If possible, I would prefer that we use this version.

I can add type annotations if you want, yes. 

>(i.e. just a side effect or really a value that we don't care about).

I don't understand this distinction though. If ignore is needed then this necessarily means that we have a function that returns a value that we do not care about and that we want to throw away.

UnixJunkie:
I meant: ignoring a unit is usually very different than ignoring anything else.
You ignore a unit because you did not write begin ... end.
You ignore any other value because you do something "strange", usually.

rixed:
> I am not sure about @rixed 's plan for the v3 branch.

I have no plan regarding v3. I think we should release the next 2.x version, then rebase v3 onto master, then fix all tests, then release v3. I can help with some of that if needed.

Issue #929, opened 2019-11-28
Unix.sleepf


UnixJunkie:
Could be provided across more OCaml versions, since it can be implemented using Unix.select
in case it is not provided by the stdlib.

UnixJunkie:
I have to admit that I added a delay of at least 2ms to the regression test suite.
My apologies.

UnixJunkie:
was merged

PR #930, opened 2019-11-28
make Unix.sleepf available under all OCaml versions

UnixJunkie:


UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/929

src/batUnix.mlv
```diff
@@ -30,6 +30,27 @@ include Unix
 ##V<4.2##let send_substring = send
 ##V<4.2##let sendto_substring = sendto
 
+##V<4.3##let sleepf (timeout: float): unit =
+##V<4.3##  let _, _, _ = select [] [] [] timeout in
```
rixed: You probably want to restart_on_EINTR.

Also, have you considered `Thread.delay` ?

UnixJunkie: I'll have a look at that. Thanks for the review.

UnixJunkie: Concerning thread.delay, it is implemented as Unix.sleepf in recent OCaml versions.

UnixJunkie: the implementation became a little bit more complex

UnixJunkie:
I also updated the changelog.

src/batUnix.mlv
rixed: I'm assuming `Thread.delay` does not use `Unix.sleepf` when V<4.3, so wouldn't that be easier and more reliable to just call `Thread.delay` then? Unless we do not want to link the Unix to the Thread module, maybe?

UnixJunkie:
If you have a better PR, or if you can improve this one, please do it.

src/batUnix.mlv
UnixJunkie: They use thread_delay from scheduler.c, which calls schedule_thread, which looks like a real mess.
But, is the semantics of delaying a thread the same as making a process sleep?
I think a process could have several threads, in theory.

Issue #938, opened 2020-01-13
BatResult.t = Stdlib.result


jrochel:
Shouldn't there be a type equality ? Otherwise one is forced to continually convert between the types.

UnixJunkie:
Do you have a PR to propose?

jrochel:
I've become quite reticent concerning PRs to any OCaml packages as the community turns out to be what I would consider extremely conservative to change. This is why I have made it a habit feel to out the terrain first (as I'm doing with this issue).

UnixJunkie:
I agree that OCaml's stdlib is "extremely conservative" and same for several venerable OCaml hackers. However, batteries is not the stdlib.
If there is a technical fix for the problem you observe, I would be happy to see the fix you propose; hence my request for a PR.

UnixJunkie:
A PR from a helpful contributor was merged.

PR #939, opened 2020-01-16
Add equality between Stdlib.result and BatInnerPervasives.result

facelesspanda:
See #938 

I agreed with the point raised by the issue author. I don't think more changes are needed. If I missed some places to put tests, I'd like to know.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -46,6 +46,10 @@ Changelog
   are changed. The default value is Pervasives.compare.
   #679
   (Cedric Cellier)
+- BatResult: BatPervasives.result is now equal to Stdlib.result
+  instead of sharing constructors without being the same type
+  #938
+  (Clément Busschaert)
 
 ## v2.11.0 (minor release)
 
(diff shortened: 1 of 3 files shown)


rixed:
According to OCaml "Changes" file, the result type appeared in OCaml v4.08.

facelesspanda:
okay, I see. Makes sense. Thanks, I'll do that.

UnixJunkie:
A possible test is to opam switch to 4.07 and compile batteries, then switch to 4.08 and recompile.
I am not sure if it is possible to unit-test this.

hcarty:
The `result` type appeared in 4.03. A `Result` module was added in 4.08.

facelesspanda:
so, no need to change the type definition?
Though I may have to remove the compatibility with `Legacy.Result`, right? because before 4.07 that module doesn't exist?

UnixJunkie:
For things which are version-dependant, you can use preprocessor directives, as stated before by rixed.

facelesspanda:
As a matter of fact, the compatibility with the `Stdlib.Result` module fails because `Stdlib.Seq.t` is not `BatSeq.t`. I'll look into that too.

rixed:
Superseeded by https://github.com/ocaml-batteries-team/batteries-included/pull/957

Issue #940, opened 2020-02-03
feature request : Add find_option in BatMap.S interface


nicoTolly:
Currently the module type BatMap.S does not expose the function find_opt: key -> 'a t -> 'a option
which I find convenient sometimes, even if it is easy to implement it as :
let find_opt k m =
try Some (find k m) with
| NotFound -> None

Actually such a function is already implemented as find_option at line 174 of file BatMap.ml (the equivalent function in List is called find_opt). Is there any reason not to expose it ? I can do a PR if this seems ok.

gasche:
Having the function sounds like a good idea. It should be named `find_opt`, to use the same name as the standard library `Map.S` (where `find_opt` is present since OCaml 4.05). Note that the implementation you propose can be improved by using `match .. with exception`:

```
match find k m with
| exception Not_found -> None
| v -> Some v
```

(Here the difference is somewhat minor, but it is clearer in terms of which part of the code precisely may raise the exception.)

UnixJunkie:
I think this code will not compile with old versions of OCaml.

nicoTolly:
What is the policy with regards to backward compatibility ? Which versions of the compiler should I try to run on ? I can't find anything in https://github.com/ocaml-batteries-team/batteries-included/wiki/Developers-Guidelines

UnixJunkie:
Good question.
I think batteries can still compile with 3.12.1 nowadays.
But, maybe the maintainers and/or the batteries users might decide a version number
higher than this as a hard constraint.

UnixJunkie:
For example, the opam package mentions OCaml >= 4.0.0

nicoTolly:
So I did a PR, waiting for your comments (especially on the syntax matter).

UnixJunkie:
PR was merged.
I should updated the ChangeLog...

UnixJunkie:
If you want to be called something else than nicoTolly in the ChangeLog, please let use know.

PR #941, opened 2020-02-04
Implemented find_opt in Map.S interface

nicoTolly:
Related to this issue : https://github.com/ocaml-batteries-team/batteries-included/issues/940

This Pull request adds a find_opt function to the Map.S interface. It appears that there was  already a find_option implemented in Map.Concrete so I just used that. Another implementation was needed in BatSplay.ml so I added one. Also note that a function with the same type is implemented in Map.S.Exceptionless. Maybe we should do some factorization, but for now I changed the code as little as possible.
There was a discussion on the issue on which syntax should be used (about whether or not use the match... exception syntax). I'm still not clear on the subject, so I'm waiting for your comments.

### DIFF
--- a/src/batMap.ml
+++ b/src/batMap.ml
@@ -964,7 +964,6 @@ let add x d m = Concrete.add x d Pervasives.compare m
 let update k1 k2 v2 m = Concrete.update k1 k2 v2 Pervasives.compare m
 
 let find x m = Concrete.find x Pervasives.compare m
-let find_opt x m = Concrete.find_option x Pervasives.compare m
 
 (*$T add; find
   empty |> add 1 true |> add 2 false |> find 1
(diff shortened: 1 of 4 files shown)


UnixJunkie:
Thanks a lot for sending a PR.

Can you replace the implementation by this:
```
let find_opt k m =
  try Some (find k m)
  with Not_found -> None
```
Because it will compile whatever the OCaml version.

Also, the unit tests are usually a few line as comments in the same file.
You don't need to change test_map.ml.

nicoTolly:
Done. As an aside, it seems that make test does not automatically recompile every changed files (at some point, make test failed, but make clean/make test succeeded without changing the code). Is it a known problem ?

UnixJunkie:
Thanks a lot, perfect!

UnixJunkie:
I think I noticed this recompile problem. But, I am too lazy to dig into a hairy Makefile.
Also, we should use dune and get rid of the current build system in the long run, which
also adds to my low motivation...

rixed:
FTR: these days dependencies are not handled with make but with ocamlbuild

rixed:
find_opt has been added only in v4.05 of ocaml stdlib whereas batteries claim to support ocaml >= 4.00, therefore compilation is broken for those older compiler.

PR #936, opened 2019-12-30
Add functions added in OCaml 4.10

kit-ty-kate:
Fixes https://github.com/ocaml-batteries-team/batteries-included/issues/935.

This can be viewed as a first step, some of the functions can be reproduced in batteries to avoid them being available only with OCaml 4.10

PR #943, opened 2020-02-17
Fix tests on OCaml 4.10

kit-ty-kate:
Fixes https://github.com/ocaml-batteries-team/batteries-included/pull/936#issuecomment-586996736 and a bunch of failures that came after.

My first PR seemed to have been a bit too rough. To keep compatibility with the standard library I removed `BatList.find_map_opt` in favour of changing the type of `BatList.find_map`. One other solution is to set that in the tests but I figured since it's a major release it might be worth doing it. Otherwise I can always do that instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ env:
   - OCAML_VERSION=4.06.0
   - OCAML_VERSION=4.07.0
   - OCAML_VERSION=4.08.0
+  - OCAML_VERSION=4.09.0
+  - OCAML_VERSION=4.10.0+rc2 OCAML_BETA=enable
 
 # notifications:
 #   email:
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
The alternative fix for find_map would be: https://github.com/ocaml-batteries-team/batteries-included/commit/e2798e76439ccd844e48c49d9b9a6297aed54309

Tell me what you prefer I have a branch ready with the alternative fix: https://github.com/kit-ty-kate/batteries-included/tree/fix-410-alt

kit-ty-kate:
cc @rixed

UnixJunkie:
I think this can be closed.
I don't have a 4.10.0 switch installed to test, though.

PR #944, opened 2020-02-17
Fix tests on OCaml 4.10 (alt fix)

kit-ty-kate:
Opening this one just in case to see more clearly the changes between https://github.com/ocaml-batteries-team/batteries-included/pull/943 and this

cc @rixed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ env:
   - OCAML_VERSION=4.06.0
   - OCAML_VERSION=4.07.0
   - OCAML_VERSION=4.08.0
+  - OCAML_VERSION=4.09.0
+  - OCAML_VERSION=4.10.0+rc2 OCAML_BETA=enable
 
 # notifications:
 #   email:
--- a/src/batPrintexc.mliv
+++ b/src/batPrintexc.mliv
@@ -170,6 +170,7 @@ val print : _ BatInnerIO.output -> exn -> unit
 ##V>=4.2##(** {6 Current call stack} *)
 ##V>=4.2##
 ##V>=4.2##val get_callstack: int -> raw_backtrace
+##V>=4.10##external get_callstack : int -> raw_backtrace = "caml_get_current_callstack"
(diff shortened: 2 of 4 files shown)


kit-ty-kate:
Is travis not enabled?

UnixJunkie:
The travis thing is broken for quite some time.

kit-ty-kate:
This should still be merged though

UnixJunkie:
I thought @rixed fixed the tests for 4.10 very recently...

rixed:
On current master (aka without this fix):
```
$ make test
...
File "src/batteries_compattest.ml", line 38, characters 17-21:
38 |   module List = (List: module type of Legacy.List)
                      ^^^^
Error: Signature mismatch:
       ...
       Values do not match:
         val find_map : ('a -> 'b option) -> 'a list -> 'b
       is not included in
         val find_map : ('a -> 'b option) -> 'a list -> 'b option
       File "list.mli", line 240, characters 0-55: Expected declaration
       File "src/batList.mliv", line 461, characters 0-49: Actual declaration
Command exited with code 2.
Compilation unsuccessful after building 194 targets (192 cached) in 00:00:00.
make: *** [Makefile:252: test-compat] Error 10
```

With the fix:
```
$ git merge kit-ty-kate/fix-410-alt
(...)
$ make test
Ran: 234 tests in: 6.45 seconds.
OK
```

So this patch is indeed fixing this issue.

UnixJunkie:
thanks for the test!

rixed:
Shall we move the v3.0.0 tag then?

kit-ty-kate:
> Shall we move the v3.0.0 tag then?

That will have to be `v3.0.1`. `v3.0.0` has already been released in opam-repository.

rixed:
Ah OK I was unaware of this.

UnixJunkie:
Since it is just a fix for tests, as far as I understand, I don't think we need to issue a new tag and release again.

Issue #951, opened 2020-04-10
BatFile.nb_lines


UnixJunkie:
Quite often, when system programming while riding my caml,
I need to count the number of lines in a file.
I am tired of writing variations of this time and again, so I will send an implementation here.

UnixJunkie:
Maybe count_lines would be a better name.

UnixJunkie:
If someone could check that this does the right thing on windows,
that would be helpful:
```
let count_lines (fn: string): int =
  let count = ref 0 in
  let input = open_in fn in
  (try
     while true do
       let _line = input_line input in
       incr count
     done
   with End_of_file -> close_in input);
  !count
```

rixed:
Indeed, `nb_lines` does not sound right to me. Actually `nb` may be not be very common outside France (see for instance https://www.proz.com/kudoz/french-to-english/general-conversation-greetings-letters/780428-nb-%3D-number.html for an illustration of how it can puzzle even translators).
I think `count` or `num` would be preferable.

Now, why not use one of the 3 (!) `lines_of` function that we already have? So that `count_lines` would be along the lines of `Enum.hard_count % File.lines_of` ?

UnixJunkie:
In my experience, BatIO is inefficient compare to its stdlib counterparts.

UnixJunkie:
was merged

PR #953, opened 2020-04-13
added BatFile.count_lines

UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/951

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,10 @@ Changelog
 
 ## NEXT_RELEASE (minor release)
 
+- BatFile: added count_lines
+  #XXX
+  (Francois Berenger, review by YYY)
+
 - BatArray: use unsage_get and unsafe_set more often
   #947
   (Francois Berenger, review by Cedric Cellier)
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,7 +4,7 @@ Changelog
 ## NEXT_RELEASE (minor release)
 
 - BatFile: added count_lines
-  #XXX
+  #953
   (Francois Berenger, review by YYY)
 
 - BatArray: use unsage_get and unsafe_set more often
--- a/src/batFile.ml
+++ b/src/batFile.ml
@@ -18,6 +18,23 @@
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  *)
 
+(* this code is purposedly before any module open directive *)
+let count_lines (fn: string): int =
+  let count = ref 0 in
+  let input = open_in fn in
+  (try
+     while true do
+       let _line = input_line input in
+       incr count
+     done
+   with End_of_file -> close_in input);
+  !count
+
+(*$T count_lines
+  (Sys.file_exists "/dev/null") && (count_lines "/dev/null" = 0)
+  (Sys.file_exists "/etc/services") && (count_lines "/etc/services" > 0)
+*)
+
 open BatIO
 open ListLabels
 open Unix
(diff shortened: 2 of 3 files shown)


rixed:
You said in the previous PR that File.lines_of was slow and indeed there is a benchmark for lines_of in benchsuite (which build I just fixed in a PR), and the results are indeed not encouraging:

```
...
For reading setup.ml into a list, 
readfile (585.19 us) is 21.5% faster than
file_lines_of (745.21 us) which is 28.2% faster than
lines_of2 (1.04 ms) which is 40.7% faster than
push_lines_of (1.75 ms) which is 2.0% faster than
push_lines_of2 (1.79 ms) which is 55.4% faster than
readfile_batteries (4.01 ms)
```

where `readline` is a function similar to yours, and `readfile_batteries` is using File.lines_of :)
So it naturally raises the question: Given this abysmal performance why not replace the super slow `File.lines_of` that's based on BatIO with the one based on Stdlib that's in the benchmark?

UnixJunkie:
Please do so, but I think this is a separate issue.
Unless you can accelerate the count_lines proposed here.

UnixJunkie:
@rixed can I merge this one?
The code is indeed very close to the readfile from the bench.

rixed:
I thought you would want to turn your function into a readfile and then use that instead, but that's no problem; we can as well proceed in the other way around: you merge this PR, then I add a readfile and see if I can use it to refactor count_lines without impeding performances.

UnixJunkie:
Note that on a very large file, you don't want to have all the lines staying in memory
before counting them all (they will probably not fit).

UnixJunkie:
Maybe, if we want a generic interface, we should provide a BatFile.{iter_lines,map_lines,fold_lines}.
Then count_lines' code could becomes simpler while probably as efficient.
I work a lot with text files, and I always have such functions in a Utils module.

vbgl:
Why is `/etc/services` used in the test case?

I’ve got a configuration in which the test fails because `Sys.file_exists "/etc/services"` is `false` (i.e., this test does not test the `count_lines` function only).

vbgl:
To be more constructive, here is a possible patch:

~~~
diff --git a/src/batFile.ml b/src/batFile.ml
index 04e32df7..2be3cf77 100644
--- a/src/batFile.ml
+++ b/src/batFile.ml
@@ -32,7 +32,7 @@ let count_lines (fn: string): int =
 
 (*$T count_lines
   (Sys.file_exists "/dev/null") && (count_lines "/dev/null" = 0)
-  (Sys.file_exists "/etc/services") && (count_lines "/etc/services" > 0)
+  (Sys.file_exists __FILE__) && (count_lines __FILE__ > 0)
 *)
 
 open BatIO
~~~

UnixJunkie:
@vbgl can you send a PR?

UnixJunkie:
What is this __FILE__ variable you are using?
/etc/services is a file which should be present on most unices.

UnixJunkie:
@vbgl maybe you would prefer this one?
```
diff --git a/src/batFile.ml b/src/batFile.ml
index 04e32df7..1c67ce1c 100644
--- a/src/batFile.ml
+++ b/src/batFile.ml
@@ -31,8 +31,8 @@ let count_lines (fn: string): int =
   !count
 
 (*$T count_lines
-  (Sys.file_exists "/dev/null") && (count_lines "/dev/null" = 0)
-  (Sys.file_exists "/etc/services") && (count_lines "/etc/services" > 0)
+  (not (Sys.file_exists "/dev/null")) || (count_lines "/dev/null" = 0)
+  (not (Sys.file_exists "/etc/services")) || (count_lines "/etc/services" > 0)
 *)
 
 open BatIO
```

vbgl:
<https://ocaml.org/releases/4.10/htmlman/libref/Stdlib.html#VAL__FILE__>

I’m not really competent about unit testing, but your suggestion looks fishy.

— Is the `count_lines` functions correct?
— Of course, I don’t have any "/etc/services" file around!

PR #962, opened 2020-07-24
Adapt to OCaml 4.11

vouillon:
`make test` passes with both OCaml 4.10.0 and OCaml 4.11.0+beta2


UnixJunkie:
'make test' already works with for 4.10.0, according to my own test.

UnixJunkie:
thanks for the many fixes for 4.11

UnixJunkie:
the official instructions to install this compiler don't work, so I am not testing further:
```
# opam --version
2.0.3
# opam switch create ocaml-variants.4.11.0+beta2 --repositories=default,beta=git+https://github.com/ocaml/ocaml-beta-repository.git
[ERROR] No compiler matching 'ocaml-variants.4.11.0+beta2' found, use 'opam
        switch list-available' to see what is available, or use '--packages' to
        select packages explicitly.
```

UnixJunkie:
Ok, I could fix the problem with my opam that was not accepting to install any new switch.

UnixJunkie:
thanks a lot for the fixes

PR #964, opened 2020-08-27
BatString: Fix doc section on splitting

rixed:
itself was split by iterators section.

The whole documentation organisation into sections needs some live, though.


UnixJunkie:
thanks!

Issue #965, opened 2020-09-24
BatOptParse: %prog ist replaced only once


OnkV:
* Batteries 3.1.0
* OCaml 4.11.1
* Debian 10

The documentation for `BatOptParse.OptParser` promises:

```
val make : ?⁠usage:string -> ?⁠description:string -> ?⁠version:string
    -> ?⁠suppress_usage:bool -> ?⁠suppress_help:bool
    -> ?⁠only_leading_opts:bool -> ?⁠prog:string
    -> ?⁠formatter:Formatter.t -> unit -> t
Creates a new option parser with the given options.

parameter usage
  Usage message. The default is a reasonable usage message for most programs.
  Any occurrence of the substring "%prog" in usage is replaced with the name of the
  program (see prog).
[...]
```

But only the first occurrence of `%prog` is substituted.

See file ` batteries-included/src/batOptParse.ml`:

```
let unprogify optparser s =
    (snd (BatString.replace ~str:s ~sub:"%prog" ~by:optparser.op_prog))
```

`BatString.replace` only replaces the first occurrence of the substitution string.

IMHO either the documentatioin or the implementation has to receive an update.  Thank you!

UnixJunkie:
Can you send a fix?

UnixJunkie:
It's pretty lame, but we don't even have a BatString.replace_all...

UnixJunkie:
Ok, replace_all is named nreplace.

UnixJunkie:
I sent a PR so that the code does what the documentation says.
If any maintainer is not OK with that, please raise your hand soon.

UnixJunkie:
fixed by #966 
The ChangeLog needs to be updated.

PR #966, opened 2020-09-25
BatOptParse.unprogify fix

UnixJunkie:
Any occurrence of the substring "%prog" in usage is replaced with the name of
the program. Previously, only the first occurrence was replaced.


UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/965

PR #970, opened 2020-10-22
Provide an explicitly unthreaded package

vouillon:
When using Dune, there is currently no way to use batteries without threads since Dune always sets the mt flag. This provides an alternative `batteries.unthreaded` package which does not depend on the `threads` package.


gasche:
Looks good, thank you!

gasche:
It would be nice to eventually have a release with this "packaging fix". ( cc @UnixJunkie, @rixed : I don't know who is in the mood for releasing Batteries these days. )

UnixJunkie:
Maybe next week I'll have some time.

Issue #893, opened 2019-04-23
I sugest we add fold_while to BatArray as well


UnixJunkie:
We have it in batList now; but I would need it for arrays too.
I will send a PR so that others can review it.
I will try to use the same signature than BatList.fold_while.
We'll see.

UnixJunkie:
I propose this signature:
```
val fold_while :
  ('acc -> 'a -> bool) ->
  ('acc -> 'a -> 'acc) -> 'acc -> 'a mappable -> 'acc * int
```
The returned int is to tell the caller the index of the first element for which the predicate failed.

UnixJunkie:
As a reminder, here is the signature for BatList.fold_while:
```
val fold_while : ('acc -> 'a -> bool) -> ('acc -> 'a -> 'acc) -> 'acc -> 'a list -> 'acc * 'a list
```

UnixJunkie:
PR was merged

PR #974, opened 2020-10-30
Array fold while

UnixJunkie:


UnixJunkie:
related to  https://github.com/ocaml-batteries-team/batteries-included/issues/893

UnixJunkie:
Closes #893

rixed:
LGTM, but for the commits: I think it would be nicer to have first a cosmetic commit with all the "Array." module prefix removal, and then a distinct one for the addition of fold_while itself.

Issue #502, opened 2014-01-07
shouldn't we have fold as an alias for fold_left?


UnixJunkie:
For example in BatList or some Foldable interface if there is one.
Usually, when I fold, it is always from the left.

UnixJunkie:
merged both PRs

PR #976, opened 2020-10-30
BatArray.fold is an alias for fold_left

UnixJunkie:


UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/502

PR #977, opened 2020-10-30
Expose Map.remove_exn and Set.remove_exn

### LINKED ISSUE #954: Expose `Map.remove_exn`
Currently the only function that removes an element from a Map and let the caller know if that element was indeed in the Map is `extract`, that also returns the removed element, which might not be needed.

I've seen cases where remove_exc is exactly what was needed. It is already defined, just not declared. Given we already have many visible `_exc` variant of functions dealing with containers, I think it makes sense to also expose Map.remove_exc.

rixed:
Closes #954

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,10 @@ Changelog
   #966
   (Francois Berenger, report by OnkV)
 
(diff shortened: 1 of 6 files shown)


UnixJunkie:
thanks for the PR!

PR #978, opened 2020-10-30
Aliases for List.unfold_exc and RefList.find_exc

rixed:
The_exn suffix is favored.


UnixJunkie:
I think you have removed unfold_exc from the batList.mliv file.
We don't want to remove something that already exists for compatibility reasons.

rixed:
Oh you are right, give me 10mins please to fix this.

UnixJunkie:
thanks for the PR

PR #980, opened 2020-11-05
Add support for OCaml 4.12

kit-ty-kate:
Compat tests don't pass but it can be done in two phases (two different releases)


UnixJunkie:
Do you want to push more things in there or can it be merged already?

kit-ty-kate:
This was an unforseen runtime failure fix. I haven't seen any other runtime failures so hopefully this should be good to merge

UnixJunkie:
thanks!

PR #981, opened 2020-11-13
Removed double cut_on_char that appears twice in the BatString interface

fccm:
https://github.com/ocaml-batteries-team/batteries-included/blob/8db9a29365b31d689677565705154dc5b6595020/src/batString.mliv#L795


UnixJunkie:
nice catch

PR #983, opened 2020-11-13
Removed double trim in BatString's interface

fccm:
- trim appears twice in batString.mliv
- and added char '\026' in the comments


UnixJunkie:
thanks!

PR #985, opened 2020-11-15
Add Feature parity of DynArray and Array in the ChangeLog

fccm:


UnixJunkie:
Thanks, I updated it: the feature parity is not complete (but don't bother, this module is not very heavily used I think).

PR #987, opened 2020-11-15
DynArray.{left,right,tail} need user input checks

fccm:
Dangerous functions if we don't check user input, because unsafe assignments are used.


UnixJunkie:
Thanks for the improvements.
The corresponding ocamldoc could mention which function throws which exception and under what conditions.

PR #990, opened 2020-11-27
Fix Int32/64.pow for negative exponents

### LINKED ISSUE #989: Stack overflow on Int{32,64}.pow with negative exponent
```ocaml
# Int32.(pow 1l (of_int ~-1));;
Stack overflow during evaluation (looping recursion?).
Raised by primitive operation at file "src/batNumber.ml", line 279, characters 14-31
```

Will submit a patch.

rixed:
Closes #989

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,10 +9,9 @@ Changelog
   #872
   (andrepd, review by Florent Monnier and Francois Berenger)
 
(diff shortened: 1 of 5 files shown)


UnixJunkie:
You don't want to be able to compute 2^(-3) ?
It seems valid to me.

UnixJunkie:
Ok, because the return type is int.

PR #991, opened 2020-12-03
Add support for ocaml-multicore

kit-ty-kate:
ocaml-multicore (OCaml 5.0) changes the interface for the `Gc` module, the interface is not set in stone yet so this PR is not intended to be merged now, only as a way for people testing ocaml-multicore early to be able to use packages that depend on batteries.

### DIFF
--- a/build/prefilter.ml
+++ b/build/prefilter.ml
@@ -1,6 +1,6 @@
-let (major, minor) =
+let (major, minor, extra) =
   Scanf.sscanf Sys.ocaml_version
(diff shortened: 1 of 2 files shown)


gasche:
Why not merge it now? It should not break any code as the new logic does not fire on existing OCaml versions, so there is no downside for non-multicore users, and the worst-case for multicore users is that the logic you wrote is not quite right, so they get a build error (as today).

kit-ty-kate:
Fair enough, I just marked it a ready for review. Feel free to merge.

Issue #995, opened 2020-12-17
BatArray.remove_at


UnixJunkie:
Would be the reverse of the existing:
```
val BatArray.insert: 'a array -> 'a -> int -> 'a array
```

UnixJunkie:
I propose:
```
val BatArray.remove: 'a array -> int -> 'a * 'a array
```

UnixJunkie:
I had to code it yesterday...
It annoys me when I encounter small functionalities that batteries doesn't have.

fccm:
has been merged with PR #996
Thanks!

PR #996, opened 2020-12-18
BatArray.remove_at

UnixJunkie:


UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/995

src/batArray.mlv
```diff
@@ -822,6 +822,22 @@ let insert xs x i =
      with Invalid_argument _ -> true
 *)
 
+let remove_at i src =
+  let x = src.(i) in (* keep the bound check in there *)
+  let n = length src in
+  let dst = make (n - 1) x in
```
rixed, line 828: For the record: In theory, Array.init would permit to initialize `dst` only once taher than twice (first with `x`, second with actual values). In practice IIRC it's the same under the hood.

UnixJunkie: I thought about this.
If it's faster, why not.
I think it might also make the code more readable.
Thanks for the review by the way.

PR #998, opened 2020-12-29
Little bit faster BatList.cartesian_product

fccm:
More details in this issue: #997

Also by avoiding the call to [List.concat](https://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html#VALconcat) seems to make this function tail-rec, while the doc says that [List.concat](https://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html#VALconcat) is not tail-rec.

You can test this PR with:
```
$ ocamlfind ocamlopt -linkpkg -package "unix,num,bigarray,str,bytes" \
        -I _build/src/ batteries.cmxa -o cp.opt cp.ml

$ cat cp.ml

let cartesian_product_orig l1 l2 =
  List.concat (List.map (fun a -> List.map (fun b -> (a, b)) l2) l1)

let cartesian_product = BatList.cartesian_product
let cartesian_product_rev = BatList.cartesian_product_rev

let () =
  let l1 = List.init 2_000 (fun i -> i) in
  let l2 = List.init 4_000 (fun i -> i) in
  let p =
    match Sys.argv.(1) with
    | "concat" -> cartesian_product_orig l1 l2
    | "fold" -> cartesian_product l1 l2
    | "fold_rev" -> cartesian_product_rev l1 l2
    | _ -> raise Exit
  in
  Printf.printf "len: %d\n%!" (List.length p);
;;

$ time ./cp.opt concat
len: 8000000

real	0m2,566s
user	0m2,392s
sys	0m0,172s

$ time ./cp.opt fold
len: 8000000

real	0m2,355s
user	0m2,215s
sys	0m0,139s

$ time ./cp.opt fold_rev
len: 8000000

real	0m2,176s
user	0m2,046s
sys	0m0,128s
```

PR #999, opened 2020-12-30
Provide BatList.cartesian_product_(iter|fold)

fccm:
You should merge #998 first before this one.

### DIFF
--- a/src/batList.mliv
+++ b/src/batList.mliv
@@ -919,6 +919,20 @@ val cartesian_product_rev : 'a list -> 'b list -> ('a * 'b) list
 
     @since NEXT_RELEASE *)
 
+val cartesian_product_iter : ('a * 'b -> unit) -> 'a list -> 'b list -> unit
+(** [cartesian_product_iter f l1 l2] passes to the function [f] every possible
+    pairs from elements from the two input lists.
+
+    @since NEXT_RELEASE *)
+
+val cartesian_product_fold : ('c -> 'a * 'b -> 'c) -> 'c -> 'a list -> 'b list -> 'c
+(** [cartesian_product_fold f init l1 l2] is similar than [List.fold_left]
+    but passes to the function [f] every possible pairs of elements from the
+    two input lists [l1] and [l2], and using an accumulator initialised with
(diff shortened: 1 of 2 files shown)


UnixJunkie:
I personally don't vouch for this PR.
Other maintainers are welcome to come and rescue it if they feel like so.

fccm:
OK, I can still copy it in my own projects.
I thought it could maybe be useful, while in *game-dev* these are iterators that are very and often useful, like for example to check interactions between every enemies with every player bullets.
But never mind, someone looking for it can find these in containers.

gasche:
What is the performance difference between

```
List.cartesian_product_iter (fun a b -> ...) la lb
```

and 

```
la |> List.iter @@ fun a ->
lb |> List.iter @@ fun b ->
....
```

?

UnixJunkie:
@fccm all against all tests are very costly. You should probably use some smarter collision detection data-structure / algorithms.
Usually, this topic is covered in computer graphics books, e.g. because of needs in ray-tracing.

gasche:
To rephrase my point: my impression is that `cartesian_iter` and `cartesian_fold` are unnecessary because they can be simply written as two nested iterations on the lists. Having a function to compute the cartesian product is difficult to express as a tail-recursive computation, so it makes sense to provide it. This being said, in the `_iter` and `_fold` cases I have not actually measured the different, so I might be wrong; I think the onus to do this would be on @fccm, and he may even find (through this experiment) that he himself doesn't need those functions after all.

fccm:
> To rephrase my point: my impression is that `cartesian_iter` and `cartesian_fold` are unnecessary because they can be simply written as two nested iterations on the lists. Having a function to compute the cartesian product is difficult to express as a tail-recursive computation, so it makes sense to provide it. This being said, in the `_iter` and `_fold` cases I have not actually measured the difference, so I might be wrong; I think the onus to do this would be on @fccm, and he may even find (through this experiment) that he himself doesn't need those functions after all.

Yes, you're right. It's just that nesting two `List.fold_left` is not that easy to figure out for a beginner, that sometimes even find difficult to use a single `List.fold_left`.

PR #1000, opened 2021-01-05
Map findfirst

jabot2:
adds find_first, find_first_opt, find_last and find_last_opt methods, compatible with the stdlib.

### DIFF
--- a/src/batMap.ml
+++ b/src/batMap.ml
@@ -181,6 +181,15 @@ module Concrete = struct
       | Empty -> found in
     loop None map
 
+  (*$T find_first_opt
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_first_opt (fun x -> x >= 0)) = (Some (1, 11))
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_first_opt (fun x -> x >= 1)) = (Some (1, 11))
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_first_opt (fun x -> x >= 2)) = (Some (2, 12))
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_first_opt (fun x -> x >= 3)) = (Some (3, 13))
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_first_opt (fun x -> x >= 4)) = (None)
+    (empty |>                                     find_first_opt (fun x -> x >= 3)) = (None)
+  *)
+
   let find_first f map =
     match find_first_opt f map with
     | Some x -> x
@@ -196,6 +205,15 @@ module Concrete = struct
       | Empty -> found in
     loop None map
             
+  (*$T find_last_opt
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_last_opt (fun x -> x <= 0)) = None
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_last_opt (fun x -> x <= 1)) = Some (1, 11)
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_last_opt (fun x -> x <= 2)) = Some (2, 12)
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_last_opt (fun x -> x <= 3)) = Some (3, 13)
+    (empty |> add 1 11 |> add 2 12 |> add 3 13 |> find_last_opt (fun x -> x <= 4)) = Some (3, 13)
+    (empty |>                                     find_last_opt (fun x -> x <= 3)) = None
+  *)
+
   let find_last f map =
     match find_last_opt f map with
     | Some x -> x
@@ -1030,7 +1048,12 @@ let find_default def x m =
 (*$T find_default
     find_default 3 4 (add 1 2 empty) = 3
     find_default 3 1 (add 1 2 empty) = 2
-*)
+ *)
+
+let find_first f map = Concrete.find_first f map
+let find_first_opt f map = Concrete.find_first_opt f map
+let find_last f map = Concrete.find_last f map
+let find_last_opt f map = Concrete.find_last_opt f map
 
 (*$T pop_min_binding
   (empty |> add 1 true |> pop_min_binding) = ((1, true), empty)
@@ -1249,6 +1272,12 @@ module PMap = struct (*$< PMap *)
     find_default 3 1 (add 1 2 empty) = 2
   *)
 
+  let find_first f map = Concrete.find_first f map.map
+  let find_first_opt f map = Concrete.find_first_opt f map.map
+  let find_last f map = Concrete.find_last f map.map
+  let find_last_opt f map = Concrete.find_last_opt f map.map
+
+    
   (*$T update
     add 1 false empty |> update 1 1 true |> find 1
     add 1 false empty |> update 1 2 true |> find 2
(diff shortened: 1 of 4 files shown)


UnixJunkie:
Is this a PR to make us compatible with new developments in the stdlib?

jabot2:
That was my intention.

UnixJunkie:
Is the code the exact same one from the stdlib or was it rewritten in some way?

jabot2:
I apologize for the delay.

The implementation was written from scratch.
If you prefer i can copy the stdlib implementation for BatMap; BatSplay will still need a from scratch implementation.

UnixJunkie:
Sorry for being picky. Also, I am not familiar with the internals of those two modules.
I think the implementation in the stdlib is faster (they don't accumulate using an option; they have an auxiliary
function instead).
I would prefer we do the same.
@gasche we (batteries) don't get a notification when things change/are added in the stdlib? That would be nice.

UnixJunkie:
@jabot2 when you are finished, you should update the changelog and put the name you want to leave for posterity in there

UnixJunkie:
@jabot2 Apparently, those functions were also needed in the Set module for the stdlib.

src/batSplay.ml
```diff
@@ -314,6 +314,70 @@ struct
...
+      match tr with
+      | Node(l, (k, v), r)
+        -> if f k
+           then loop (Some (k, v)) l
+           else loop found r
+      | Empty -> found in
+    let ret = loop None (sget map) in
+    (* dummy find to rebalance the tree *)
+    if Option.is_some ret
+    then ignore(find (fst (Option.get ret)) map);
```
gasche, line 328: `Option.get` guarded by `Option.is_some` is not so nice. This would be nicer with `Option.iter` or a direct pattern-matching.

gasche:
> @gasche we (batteries) don't get a notification when things change/are added in the stdlib? That would be nice.

There is `src/batteries_compattest.mlv`, but `Map` is not tested there, apparently from the start ( 8181d5f426352e371190286d29d83ab951b41050 ). Maybe we should try harder to test it as well.

src/batMap.ml
```diff
@@ -171,6 +171,36 @@ module Concrete = struct
...
 
+  let find_first_opt f map =
+    let rec loop found tree =
+      match tree with 
+      | Node(l, k, v, r, _) ->
+         if f k
+         then loop (Some (k, v)) l
+         else loop found r
+      | Empty -> found in
+    loop None map
```
gasche, line 182: If we wanted to avoid the `Some (k, v)` in the middle, we could have two founctions, `loop_not_found f tree` and `loop_found f k v tree`. (I'm including `f` as a parameter to make the functions closed, which is interesting here.)

src/batSplay.ml
```diff
@@ -314,6 +314,70 @@ struct
...
+    ret
+
+  (* option 2: no unneccessary "find" invocation (saves several comparisons),
+     but implementation is rather complicated *)
```
gasche: This implementation is too complex. I haven't looked at the details, but my intuition would be to write a `cfind_first` function.

gasche (commented):
I made some minor implementation comments inline. My most important comment is that all new functions added to Batteries must come with tests, so these new functions should have new tests.

src/batSplay.ml
jabot2: after re-writing, the find_first_opt method will call the find_first method, where the rebalancing happens after the "raise Not_found", so the Option.is_some guard is obsolete.

src/batMap.ml
jabot2: Done. The loop_found and loop_not_found functions are nested in the find_first and find_last functions. If you want i can extract them, but I think that should not be necessary.

jabot2:
@gasche : adding Map and Set to src/batteries_compattest.mlv yielded the following missing functions:
```
module Map = (Map : module type of Legacy.Map)
       ...
       The value `of_seq' is required but not provided
       The value `add_seq' is required but not provided
       The value `to_seq_from' is required but not provided
       The value `to_seq' is required but not provided
       The value `choose_opt' is required but not provided
       The value `max_binding_opt' is required but not provided
       The value `min_binding_opt' is required but not provided
       The value `union' is required but not provided

module Set = (Set : module type of Legacy.Set)
       ...
       The value `of_seq' is required but not provided
       The value `add_seq' is required but not provided
       The value `to_seq' is required but not provided
       The value `to_seq_from' is required but not provided
       The value `find_last_opt' is required but not provided
       The value `find_last' is required but not provided
       The value `find_first_opt' is required but not provided
       The value `find_first' is required but not provided
       The value `find_opt' is required but not provided
       The value `choose_opt' is required but not provided
       The value `max_elt_opt' is required but not provided
       The value `min_elt_opt' is required but not provided
```

UnixJunkie:
If @gasche agrees with the Splay part, this should be merged.

UnixJunkie:
@jabot2 can you open an issue to track those?

jabot2:
@UnixJunkie : again sorry for the delay, i had to get some sleep :-)
I will open another issue. 
With the changelog entry this PR should be done, if you don't see any more issues?

gasche:
Thanks for the revision and issue-tracking etc.! Another batch of comments:

- If you are importing implementations from the stdlib, we should (1) check that the license are compatible, and (2) credit the original author(s) of these functions. This is a legal requirement. To do (1), you have to read the LICENSE files. To do (2), my proposal would be to locate the author of those functions (I expect that there is a single contribute that wrote both functions) by using `git blame stdlib/map.mli` or `git log -u stdlib/map.mli`; then either credit them in the commit message the imports the code, or directly attribute the commit to this contributor (`git commit [--amend] --author="Foo Bar <email>"`).

- In BatSplay, you chose the simple implementation with a-posteriori rebalancing (instead of `cfind_{first,last}` functions on cursors). This is okay with me: most contributors don't actually use BatSplay, they add code there to satisfy the module signature, so let's not be perfectionist on the implementation, "simple and tested" is already very nice. But I would prefer if you avoided `Not_found` in the `find_*_opt` functions, because js_of_ocaml has slow exceptions so we should not use exceptions for control flow. (Some of the overhead can be avoided by using `raise_notrace`,  but then you have to catch that internal exception and reraise it with a backtrace in the exception-raising variant, it's not worth it when it is easy to write the code without exceptions in the first place.)

jabot2:
@gasche 
- i will amend the ChangeLog
- i will change the implementation in BatSplay. By the same logic, i will also change the find_*_opt in BatMap to not use find_* (-non-opt).

jabot2:
@gasche Also, the stdlib has LGPL v2.1, same as batteries.

src/batMap.ml
```diff
@@ -171,6 +171,99 @@ module Concrete = struct
...
+       then loop_found k v f l
+       else find_first f r
+
+  let rec find_first_opt f m =
+    let rec loop_found k0 v0 f = function
+      | Empty -> Some (k0, v0)
+      | Node (l, k, v, r, _) ->
+         if f k
+         then loop_found k v f l
+         else loop_found k0 v0 f r in
```
gasche, line 195: Unless I am missing something, the `loop_found` is identical to `loop_found` above. They could be shared by being declared at the toplevel as `find_first_before f k v m`. Same thing with the `loop_found` of the `find_last_*` functions, which could be `find_last_after f k v m`.

jabot2: good finding, thanks. I will refactor this...

gasche (approved):

gasche:
Merged, thanks for your contribution!

PR #1001, opened 2021-01-15
BatPp

UnixJunkie:
Now, even Batman has an OCaml preprocessor.

This might just do what batteries need.
I don't guarantee it is totally correct.
I just did some toy tests.

Feedback is welcome. I am not a compiler guy, please don't lough at how I parse things.
:D

### DIFF
--- a/src/batPp.ml
+++ b/src/batPp.ml
@@ -1,0 +1,117 @@
+(* batteries super simple pre processor.
+   Cannot depend on anything else than the stdlib.
+
+man ocamlc:
+  [...]
+  -pp <command>:
+    Cause the compiler to call the given command as a preprocessor
+    for  each source file. The output of command is redirected to an
+    intermediate file, which is compiled. If there are  no  compila‐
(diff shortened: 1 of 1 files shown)


UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/658

gasche:
This is already done in [build/prefilter.ml](https://github.com/ocaml-batteries-team/batteries-included/blob/ab4ada0/build/prefilter.ml).

UnixJunkie:
Mine doesn't require the str library.
I will test it more and might add some interesting features.
Then, I will submit again to batteries.

gasche:
The main problem with these preprocessors is (not the use of Str but) the fact that their directives are invalid OCaml syntax. I think it would be worth trying to change the preprocessor syntax to use OCaml comments, for example `(*#V<=4.10#*)`, instead of the current syntax `##V<=4.10##`.

PR #1004, opened 2021-01-18
better implementation of BatArray.remove_at

UnixJunkie:


UnixJunkie:
@rixed this should address the comment you had about the previous implementation

src/batArray.mlv
```diff
@@ -823,12 +823,15 @@ let insert xs x i =
...
-  let dst = make (n - 1) x in
-  blit src 0 dst 0 i;
-  blit src (i + 1) dst i (n - i - 1);
-  dst
+  if i = 0 then
+    sub src 1 (n - 1)
+  else if i = n - 1 then
+    sub src 0 (n - 1)
+  else
+    let dst = sub src 0 (n - 1) in
```
rixed, line 832: If we `remove_at 1`, most of the array will be copied twice (once in the `sub` and once in the `blit`). Isn't that a problem if the array is large?

gasche: Note: one way to avoid this would be to add a new function to the standard library `Array` module that exposes the underlying `gather` primitive implemented in C, with a type of the form `('a array * int * int) array -> 'a array`: it takes an array of "subarrays" (array, offset, length), and joins all subarrays into a single output array. This is used under the hood to implement `sub` and `concat`, but it could also be used to implement an efficient `remove_at`.

UnixJunkie:
Yes, remove_at is indeed not very efficient. People should use another data structure if they do a lot of remove_at and they want high performance. But we have it in BatList and I like that we have the same functionality in the two modules.

UnixJunkie:
Should we really open an issue against Stdlib.Array for exposing its gather C primitive?

fccm:
In the module `DynArray` the equivalent function is called `BatDynArray.delete`.
Someone who whould like to make a lot of `remove_at` operations should probably use the `DynArray.t` structure.

UnixJunkie:
I am thinking about closing this PR, because the current implementation is correct.
Only the array initialization is sub-optimal and I don't have time to try moving things in the stdlib.

PR #1005, opened 2021-01-18
added compatibility with stdlib.Seq for BatSeq

jabot2:


jabot2:
Stdlib has a Seq module, which is almost compatible with BatSeq.
This PR adds the missing methods (with testcases), adds Legacy.Seq and adds Seq to the automatic compatibility check.

src/batSeq.mlv
```diff
@@ -138,6 +138,31 @@ let of_list l =
...
+
+(*$T return
+  length (return 123) = 1
+  at (return 123) 0 = 123
+  equal (return 123) (of_list [123])
+ *)
+
+let rec unfold f u =
+  match f u with
+  | Some(v, r) -> (fun _ -> Cons(v, unfold f r))
```
gasche, line 158: `fun () ->` (whenever you can use either `_` or `()`, always prefer `()`, because sometimes it will tell you about a type mistake you are making.)

gasche (approved):
This looks nice, thanks in particular for enabling compatibility tests for `Seq` to catch future mismatches.

UnixJunkie:
@jabot2 thanks for this contribution
@gasche thanks for the review

src/batSeq.mlv
UnixJunkie: I fixed it.

PR #1006, opened 2021-01-19
Stdlib compat set

jabot2:
BatSet is missing several methods from Stdlib.Set.
Also, several existing methods in Stdlib.Set guarantee physical equality of resulting sets if possible.
This PR introduces the missing methods into BatSet, and adds the physical equality guarantees.
It also adds BatSet to the compattest.

### DIFF
--- a/src/batSet.ml
+++ b/src/batSet.ml
@@ -494,7 +494,7 @@ module Concrete = struct
 
   let filter_map cmp f e = fold (fun x acc -> match f x with Some v -> add cmp v acc | _ -> acc) e empty
 
-   let choose = min_elt (* I'd rather this chose the root, but okay *)
+  let choose = min_elt (* I'd rather this chose the root, but okay *)
   (*$= choose
     42 (empty |> add 42 |> choose)
     (empty |> add 0 |> add 1 |> choose) (empty |> add 1 |> add 0 |> choose)
@@ -995,7 +995,6 @@ module PSet = struct (*$< PSet *)
     get_cmp (create BatInt.compare) == BatInt.compare
   *)
 
-  (* TODO: ensure phys. equality for add remove update map_stdlib filter filter_map_stdlib *)
   let singleton ?(cmp = compare) x = { cmp = cmp; set = Concrete.singleton x }
   let is_empty s = Concrete.is_empty s.set
   let mem x s = Concrete.mem s.cmp x s.set
--- a/src/batSet.ml
+++ b/src/batSet.ml
@@ -1316,29 +1316,6 @@ let to_seq t =
 let to_seq_from k t =
   Concrete.to_seq_from Pervasives.compare k t
 
-(*$T add_seq
-  equal (of_list [1;2;3;4]) (add_seq (BatSeq.of_list [1;2]) (of_list [3;4]))
-  equal (of_list [3;4]) (add_seq (BatSeq.of_list []) (of_list [3;4]))
-  equal (of_list [1;2]) (add_seq (BatSeq.of_list [1;2]) (of_list []))
- *)
-
-(*$T of_seq
-  equal (of_list [1;2;3;4]) (of_seq (BatSeq.of_list [1;2;4;3]))
-  equal (of_list []) (of_seq (BatSeq.of_list []))
- *)
-
-(*$T to_seq
-  BatSeq.equal (BatSeq.of_list [1;2;3;4]) (to_seq (of_list [4;1;3;2]))
-  BatSeq.equal (BatSeq.of_list []) (to_seq (of_list []))
- *)
-
-(*$T to_seq_from
-  BatSeq.equal (BatSeq.of_list [1;2;3;4]) (to_seq_from 0 (of_list [4;1;3;2]))
-  BatSeq.equal (BatSeq.of_list [3;4]) (to_seq_from 3 (of_list [4;1;3;2]))
-  BatSeq.equal (BatSeq.of_list []) (to_seq_from 5 (of_list [4;1;3;2]))
-  BatSeq.equal (BatSeq.of_list []) (to_seq_from 5 (of_list []))
- *)
-
 (*$T subset
    subset (of_list [1;2;3]) (of_list [1;2;3;4])
    not (subset (of_list [1;2;3;5]) (of_list [1;2;3;4]))
--- a/src/batSet.ml
+++ b/src/batSet.ml
@@ -316,9 +316,6 @@ module Concrete = struct
       invalid_arg "Set.at_rank_exn i s: i >= (Set.cardinal s)"
     with Found -> !res
 
-  let map cmp f s =
-    fold (fun v acc -> add cmp (f v) acc) s empty
-
   let rec op_map f = function
     | Empty -> Empty
     | Node (l,x,r,h) -> Node (op_map f l, f x, op_map f r, h)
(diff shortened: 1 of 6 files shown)


UnixJunkie:
Thanks for this contribution.
Note that all new code must come with unit tests.

jabot2:
@UnixJunkie I thought i added unit tests for everything? I will check again...

src/batSet.ml
```diff
@@ -531,6 +679,28 @@ module Concrete = struct
...
+    Seq.fold_left (fun m e -> add cmp e m) m s
+    
+  let of_seq cmp s =
+    add_seq cmp s empty
+    
+  let rec to_seq m =
+    match m with
+    | Empty -> Seq.empty
+    | Node(l, v, r, _) -> 
+       Seq.append (to_seq l) (Seq.cons v (to_seq r))
```
gasche, line 693: Two remarks:
- This implementation does not do what I would expect due to the use of the strict function `Seq.cons`: it will traverse the whole set before returning the first element. This is not technically wrong as there is no order-of-evaluation difference (and probably not much of a performance difference, although it is worth measuring), but I would find it more natural to have a version that lazily enumerates the list.
- We have a proper datatype to represent partial enumerations of sets, `'a iter` and `cons_iter`.

I would expect something looking roughly like (untested):
```
let rec iter_to_seq = function
| E -> Seq.empty
| C (e, r, t) -> fun () -> Seq.Cons (e, iter_to_seq (cons_iter r t))

let to_seq s = iter_to_seq (cons_iter t E)
```

(`to_seq_from` also needs an adaptation.)

I would be curious to know what is the performance difference between your version, a version using `iter` as above, and also a modified version of your code that avoids `Seq.cons` to avoid forcing the tail (but otherwise traverses the set recursively in the direct way).

src/batSet.ml
```diff
@@ -794,31 +995,58 @@ module PSet = struct (*$< PSet *)
     get_cmp (create BatInt.compare) == BatInt.compare
   *)
 
-
+  (* TODO: ensure phys. equality for add remove update map_stdlib filter filter_map_stdlib *)
```
gasche, line 998: Is this TODO item still relevant? It looks like it is now done in your code. (If it's not, then it deserves a clarification of what it means.)

gasche (commented):
Thanks, this is nice work (again).

Note: for functorized modules (Set and Map) we also have non-inline tests in `testsuite/test_<module>.ml`, so that the tests can be performed on all implementations (for functions that are available everywhere). I have mixed thoughts about adding inline tests for some functions (if they could be tested this way): inline tests are more convenient, but they either force us to add redundancy (copying the tests for several interfaces) or they test less. (But there probably should have been a comment in the `.ml` file pointing at where the tests are.) I would have a slight preference for moving the tests to `testsuite/test_set.ml` when it makes sense, but I'm interested in what you think about this.

src/batSet.ml
```diff
@@ -458,6 +555,57 @@ module Concrete = struct
...
+  let try_join cmp l v r =
+    (* [join l v r] can only be called when (elements of l < v <
+         elements of r); use [try_join l v r] when this property may
+         not hold, but you hope it does hold in the common case *)
+    if (l = Empty || cmp (max_elt l) v < 0)
+       && (r = Empty || cmp v (min_elt r) < 0)
+    then join l v r
+    else union cmp l (add cmp v r)
+
+  let rec map_stdlib cmp f = function
```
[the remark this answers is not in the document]
jabot2: the map and filter_map functions from stdlib take a function elt -> elt, while the functions in Batteries take a function elt -> 'b and produce a 'b set (for the polymorphic maps)
So it seems we must keep the old versions or break compatibility with people that currently rely on map/filter_map to create sets of a different element type.

src/batSet.ml
```diff
@@ -395,15 +492,15 @@ module Concrete = struct
   let print ?(first="{") ?(last="}") ?(sep=",") print_elt out t =
     BatEnum.print ~first ~last ~sep (fun out e -> BatPrintf.fprintf out "%a" print_elt e) out (enum t)
 
-  let filter cmp f e = fold (fun x acc -> if f x then add cmp x acc else acc) e empty
-
   let filter_map cmp f e = fold (fun x acc -> match f x with Some v -> add cmp v acc | _ -> acc) e empty
 
-  let choose = min_elt (* I'd rather this chose the root, but okay *)
+   let choose = min_elt (* I'd rather this chose the root, but okay *)
```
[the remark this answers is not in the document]
jabot2: just a typo, will fix.

src/batSet.ml
jabot2: no, the comment is obsolete. I will delete it...

src/batSet.ml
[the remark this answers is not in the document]
gasche: Ah yes, thanks. I would propose to call them `map_endo` or `endo_map` (from "endo{morphism,function}", a function from a set to itself) or `homogeneous_map` rather than `map_stdlib`.

The implementation of `map_stdlib` and `filter_map_stdlib` that you propose is more efficient than the existing `map` and `filter_map` functions in the case where the mapping function preserves the order, or "nearly" preserves the order. Would you possibly be interested in using the same implementation with `try_join` for the existing/heterogeneous map functions? Note that the `==` trick cannot be reused, as it assumes that the input and output types are equal.

jabot2:
I started moving some testcases to the testsuite. I'm not done yet, sorry...

src/batSet.ml
jabot2: I modified to_seq and to_seq_from to be more lazy.
They will not do anything on invocation (except returning a closure).
Furthermore, once that closure is called, it will only go down the leftmost path in the tree to determine the first element; recursing into the right subtrees is again done in a closure.

I hope that satisfies your expectations of these functions.

src/batSet.ml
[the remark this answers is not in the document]
jabot2: To give credit where it is due: the map_stdlib and filter_map_stdlib methods are copied (mostly) verbatim from the stdlib.

I rewrote the map and filter_map functions to use try_join. I'm not convinced that makes them faster though...

src/batSet.ml
gasche: They do, thanks! I would still be curious to know whether the `'a iter` version is faster, but I don't have the time to look at this right now and maybe you are not so curious yourself, so the new versions are fine with me.

UnixJunkie:
If some tests are in another file, it would be nice to have a comment
just after each function tested in this way, saying
where the test is, so that people are not surprised that there is
no unit test in comments just after an implementation (and they know right away
where to look for it).

jabot2:
@UnixJunkie After checking again: I think i have unit tests for every function that i touched. 
@gasche I renamed the _stdlib functions to _endo

Do you think that there is anything else to do?

gasche:
Let's go ahead and merge. Thanks for the nice work.

UnixJunkie:
@jabot2 could you contribute the to_rev_seq versions?
I don't understand the code for to_seq (in batteries) and to_rev_seq (in the stdlib).

UnixJunkie:
@jabot2 this is required in batMap, batSplay and batSet for 'make test-compat' to pass.

gasche:
I can write `to_rev_seq` versions, in exchange of someone else doing the release work.

gasche:
(Ah, I had missed a recent email from @jabot2 indicated that he's interested in contributing. Of course I'm happy to let @jabot2 continue his nice contributions! It's better to broaden the pool of active contributors.)

PR #1008, opened 2021-01-30
increase compatibility between BatMap, BatSplay and stdlib.Map

jabot2:
There were several functions added to stdlib.Map, which are added in this PR.
Unfortunately, stdlib also added an update function with an incompatible type signature, 
which is why this PR adds an update_stdlib function.

For 100% compatibility, something like `  module IntMap = struct
     include BatMap.Int
     let update = update_stdlib
  end
  ` is necessary, see also the extended batteries_compattest.

Also, for several functions the exception raised on empty maps was changed to `Not_found` to increase consistency and compatibility with stdlib; the exception raised was unspecified by the documentation for these functions anyways.

### DIFF
--- a/src/batMap.ml
+++ b/src/batMap.ml
@@ -902,21 +902,21 @@ module Concrete = struct
     fun () ->
     match m with
     | Empty -> BatSeq.Nil
-    | Node(l, k, v, r, _) ->
-       if cmp k v <= 0 then
+    | Node(l, k2, v, r, _) ->
+       if cmp k k2 <= 0 then
          BatSeq.append (to_seq_from cmp k l) (fun () -> BatSeq.Cons ((k,v), to_seq r)) ()
        else
          to_seq_from cmp k r ()
-
     
   let union_stdlib cmp f m1 m2 =
     foldi
       (fun k v m ->
-        if mem k cmp m
-        then match f k v (find k cmp m) with
-             | Some v2 -> add k v2 cmp m
-             | None -> m
-        else add k v cmp m)
+        match find_option k cmp m with
+        | Some v1 ->
+           (match f k v v1 with
+            | Some vmerged -> add k vmerged cmp m
+            | None -> m)
+        | None -> add k v cmp m)
       m1
       m2
     
@@ -1514,13 +1514,22 @@ module PMap = struct (*$< PMap *)
   let is_empty x = x.map = Concrete.Empty
 
   let add x d m =
-    { m with map = Concrete.add x d m.cmp m.map }
+    let newmap = Concrete.add x d m.cmp m.map in
+    if newmap == m.map
+    then m
+    else { m with map = newmap }
 
   let update k1 k2 v2 m =
-    { m with map = Concrete.update k1 k2 v2 m.cmp m.map }
+    let newmap = Concrete.update k1 k2 v2 m.cmp m.map in
+    if newmap == m.map
+    then m
+    else { m with map = newmap }
 
   let update_stdlib k f m =
-    { m with map = Concrete.update_stdlib k f m.cmp m.map }
+    let newmap = Concrete.update_stdlib k f m.cmp m.map in
+    if newmap == m.map
+    then m
+    else { m with map = newmap }
 
   let find x m =
     Concrete.find x m.cmp m.map
@@ -1637,8 +1646,12 @@ module PMap = struct (*$< PMap *)
     Concrete.print ?first ?last ?sep ?kvsep print_k print_v out t.map
 
   let filterv  f t = { t with map = Concrete.filterv f t.map t.cmp }
-  let filter f t = { t with map = Concrete.filter f t.map t.cmp }
   let filter_map f t = { t with map = Concrete.filter_map f t.map t.cmp }
+  let filter f t =
+    let newmap =  Concrete.filter f t.map t.cmp in
+    if newmap == t.map
+    then t
+    else { t with map = newmap }
 
   let max_binding t = Concrete.max_binding t.map
   let min_binding t = Concrete.min_binding t.map
--- a/src/batMap.ml
+++ b/src/batMap.ml
@@ -108,9 +108,10 @@ module Concrete = struct
     | Node (l, _, _, _, _) -> min_binding l
     | Empty -> raise Not_found
 
-  let min_binding_opt m =
-    try Some (min_binding m)
-    with Not_found -> None
+  let rec min_binding_opt = function
+    | Node (Empty, k, v, _, _) -> Some (k, v)
+    | Node (l, _, _, _, _) -> min_binding_opt l
+    | Empty -> None
 
   let get_root = function
     | Empty -> raise Not_found
@@ -131,9 +132,10 @@ module Concrete = struct
     | Node (_, _, _, r, _) -> max_binding r
     | Empty -> raise Not_found
 
-  let max_binding_opt m =
-    try Some (max_binding m)
-    with Not_found -> None
+  let rec max_binding_opt = function
+    | Node (_, k, v, Empty, _) -> Some (k, v)
+    | Node (_, _, _, r, _) -> max_binding_opt r
+    | Empty -> None
 
   let pop_max_binding s =
     let maxi = ref (get_root s) in
--- a/src/batMap.ml
+++ b/src/batMap.ml
@@ -893,23 +893,25 @@ module Concrete = struct
   let of_seq cmp s =
     add_seq cmp s empty
 
-  let rec to_seq m =
-    fun () ->
+  let rec to_seq_hlp m =
     match m with
-    | Empty -> BatSeq.Nil
-    | Node(l, k, v, r, _) ->
-       BatSeq.append (to_seq l) (fun () -> BatSeq.Cons ((k, v), to_seq r)) ()
+    | E -> BatSeq.Nil
+    | C(k, v, r, e) ->
+       BatSeq.Cons ((k, v), fun () -> to_seq_hlp (cons_iter r e))
+      
+  let to_seq m () =
+    to_seq_hlp (cons_iter m E)
 
-  let rec to_seq_from cmp k m =
-    fun () ->
-    match m with
-    | Empty -> BatSeq.Nil
-    | Node(l, k2, v, r, _) ->
-       if cmp k k2 <= 0 then
-         BatSeq.append (to_seq_from cmp k l) (fun () -> BatSeq.Cons ((k2,v), to_seq r)) ()
-       else
-         to_seq_from cmp k r ()
-    
+  let to_seq_from cmp k m () =
+    let rec cons_enum_from k2 m e =
+      match m with
+      | Empty -> e
+      | Node (l, k, v, r, _) ->
+         if cmp k2 k <= 0
+         then cons_enum_from k2 l (C (k, v, r, e))
+         else cons_enum_from k2 r e in
+    to_seq_hlp (cons_enum_from k m E)
+
   let union_stdlib cmp f m1 m2 =
     foldi
       (fun k v m ->
--- a/src/batMap.ml
+++ b/src/batMap.ml
@@ -54,7 +54,7 @@ module Concrete = struct
   let empty = Empty
 
   let is_empty m =
-    m == Empty
+    m = Empty
 
   (* The create and bal functions are from stdlib's map.ml (3.12)
      differences from the old (extlib) implementation :
@@ -575,7 +575,7 @@ module Concrete = struct
      build the result map. The unusual parameter order was chosen to
      reflect this.  *)
   let filterv f t cmp =
-    foldi (fun k a acc -> if f a then add k a cmp acc else acc) t empty
+    foldi (fun k a acc -> if f a then acc else remove k cmp acc) t t
   let filter f t cmp =
     foldi (fun k a acc -> if f k a then acc else remove k cmp acc) t t
   let filter_map f t cmp =
@@ -913,16 +913,13 @@ module Concrete = struct
   let to_seq_from cmp k m = 
     seq_of_iter (cons_iter_from cmp k m E)
 
-  let union_stdlib cmp f m1 m2 =
-    foldi
-      (fun k v m ->
-        update_stdlib k
-          (fun v2opt ->
-            match v2opt with
-            | Some v2 -> f k v v2
-            | None -> Some v) cmp m)
-      m1
-      m2
+  let union_stdlib f cmp1 m1 cmp2 m2 =
+    let fwrap a b1 b2 =
+      match b1, b2 with
+      | Some b1, Some b2 -> f a b1 b2
+      | x, None
+        | None, x -> x in
+    heuristic_merge fwrap cmp1 m1 cmp2 m2
     
   let compare ckey cval m1 m2 =
     BatEnum.compare (fun (k1,v1) (k2,v2) -> BatOrd.bin_comp ckey k1 k2 cval v1 v2) (enum m1) (enum m2)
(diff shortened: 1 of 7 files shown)


src/batSplay.ml
```diff
@@ -769,6 +794,37 @@ struct
...
+      m
+      s
+    
+  let of_seq s =
+    add_seq s empty
+
+  let rec to_seq m =
+    fun () ->
+    (* since the tree can change shape arbitrarily, fetch all bindings and turn them into a seq. *)
+    BatSeq.of_list (bindings m) ()
```
gasche, line 810: This is not a very nice implementation. The explanation is not correct (a splay tree is a mutable pointer to a persistent tree structure, so you could just dereference the tree and then iterate on it on-demand), but more importantly the work of providing an iterator on splay tree elements has already been done for the `Enum` datatype and it is easy to reuse it. Look at the type `'a enumeration`, the function `cons_enum` and the way it is used in the functions `equal` or `compare` (they traverse two trees, you only need one, but the idea is the same).

src/batSplay.ml
```diff
@@ -769,6 +794,37 @@ struct
...
+    
+  let rec to_seq_from k m =
+    fun () ->
+    BatSeq.filter (fun (k2, _) -> Ord.compare k k2 <= 0) (to_seq m) ()
```
gasche, line 814: Again, this implementation is not satisfactory, it does not provide the expected complexity. Could you try implementing this by using a function that takes a bound `k` and a search tree `tr`, and returns an `'a enumeration` of the elements whose keys are greater or equal to `k`?

testsuite/test_map.ml
```diff
@@ -79,6 +79,9 @@ module TestMap
 
     val equal : ('a -> 'a -> bool) -> 'a m -> 'a m -> bool
 
+    (* true if add, remove, update_stdlib and filter support physical equality *)
+    val supports_phys_equality : bool
```
gasche, line 83: this is a nice trick to test this aspect of the Map module, thanks

gasche (commented):
Thanks for the code! I made a first pass of review.

I regret not insisting further to use our proper enumeration type on sets before, because the same issue is showing up again. If you are sure that you don't have the time or knowledge to provide better implementations of these functions, feel free to say so, and I can have a look, but it's actually not so hard to use them, so I think you could give it a try.

src/batMap.mli
```diff
@@ -88,16 +88,39 @@ sig
...
       This is equivalent to [add k2 v2 (remove k1) m], but more efficient
       in the case where [k1] and [k2] have the same key ordering.
+      If [k1] and [k2] have the same key ordering and [v2] is physically
+      equal to the value [k1] is bound to in [m] then the returned map will
+      be physically equal to [m]
       @raise Not_found if [k1] is not bound in [m].
-      @since 2.4.0 *)
-
+      @since 2.4.0 
+      @before NEXT_RELEASE physical equality was nor ensured.  *)
```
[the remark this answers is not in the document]
jabot2: fixed

src/batMap.mli
```diff
@@ -259,14 +288,28 @@ sig
       The returned enumeration is sorted in increasing key order. *)
 
   val min_binding : 'a t -> (key * 'a)
-  (** Return the [(key, value)] pair with the smallest key. *)
+  (** Return the [(key, value)] pair with the smallest key.
+      Raises Not_found if the map is empty.  *)
```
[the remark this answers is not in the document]
jabot2: done

src/batSplay.ml
```diff
@@ -441,7 +441,12 @@ struct
           | C (cx, Node (l, _kv, r)) -> C (cx, Node (l, (k2, v2), r))
           | C (cx, Empty) -> raise Not_found
         end
-      end
+        end
```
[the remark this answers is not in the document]
jabot2: Probably because my editor indents like this. Reverted...

src/batSplay.ml
```diff
@@ -498,7 +514,7 @@ struct
     let rec bfind = function
       | Node (Empty, kv, r) -> mini := kv; r
       | Node (l, kv, r) -> Node (bfind l, kv, r)
-      | Empty -> assert(false)
+      | Empty -> raise Not_found
```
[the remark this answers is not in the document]
jabot2: Consistency with max_binding, choose, find (_first, _last), Consistency with BatMap and compatibility with the stdlib: https://ocaml.org/releases/4.11/htmlman/libref/Map.S.html

Note that the batteries documentation does not (yet) specify which exception is raised (which, again, should be fixed by this PR), see http://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatMap.html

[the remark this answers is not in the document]
jabot2: I mean, the current situation for batMap is:
- min_binding and any raise Not_found
- pop_min_binding and pop_max_binding have assert(false)
- max_binding has invalid_arg "PMap.max_binding: empty tree"
  (the error message seems... outdated?)
None of these exceptions are documented.

For BatSplay the current situation is:
- min_binding, any and max_binding raise Not_found
- pop_min_binding and pop_max_binding have assert(false)

src/batSplay.ml
jabot2: I did not realize that it is possible to use the persistent tree structure. In hindsight it seems obvious...
I will rewrite this, i will probably copy the implementation from BatSet.to_seq.

src/batSplay.ml
jabot2: Yes, i will re-use the implementation from BatMap.

src/batSplay.ml
gasche: > i will probably copy the implementation from BatSet.to_seq or BatMap.to_seq.

We designed a datatype (`'a enumeration`) specifically to provide an efficient solution to this problem, and I am not sure why you don't want to use it.

src/batSplay.ml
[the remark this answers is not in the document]
gasche: The authors of this code used `assert false` because they believe that this point can never be reached, for any input. I haven't looked at the code deeply, but I think that the reasoning is that `choose tr` was called first, so we already know that there is at least one element. If you find this unclear, I think it would be a good idea to add a comment. If this is wrong for some reason I missed, then indeed it needs to be changed. But if it is true that this case can never be reached, then I think `assert false` is the right choice.

src/batSplay.ml
jabot2: So, i tried to implement to_seq with 'a enumeration and without:
```
  let to_seq m () = (* version with cons_enum *)
    let rec to_seq_hlp m =
      match m with
      | End -> BatSeq.Nil
      | More(k, v, r, e) ->
         BatSeq.Cons ((k, v), fun () -> to_seq_hlp (cons_enum r e)) in
    to_seq_hlp (cons_enum (sget m) End)

  let to_seq m = (*  version without cons_enum *)
    let rec to_seq_hlp m =
      fun () ->
      match m with
      | Empty -> BatSeq.Nil
      | Node(l, kv, r) ->
         BatSeq.append (to_seq_hlp l) (fun () -> BatSeq.Cons (kv, to_seq_hlp r)) () in
    to_seq_hlp (sget m)
```

I can see that `cons_enum` and the `'a enumeration` data type might be preferable to generating closures.
However, creating a `'a seq` requires creating one closure per item anyways, so we're not gaining anything.

For `to_seq_from` it gets worse: either we reimplement the `cons_enum` function to compare with the given key and skip descending into the left subtree, or we generat a full enumeration and throw away the leading N elements that are too small.

gasche: Let me first apologize for the somewhat grumpy tone of my last remark. I will look at using `'a enumeration` in BatSeq at first and see if we have a measurable performance difference, before we decide what to do for BatMap and BatSplay.

jabot2: I will try to reimplement using enumeration.

gasche: I proposed a benchmark for `BatSet.to_seq` in #1010. It seems that the enumeration/iteration-based implementation is sensibly faster. (I don't find it surprising, but it is better to have hard data than guesses.) I proposed an implementation, still for `BatSet`, in #1011; for `to_seq_from` I indeed used the approach you suggest of a new `cons_iter_from` function. Feel free to comment on the implementations, and/or to reuse the idea for Map/Splay if you find it convincing.

src/batSplay.ml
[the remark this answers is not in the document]
jabot2: You're right, i missed that get_root (for batMap) and choose (for batSplay) are called before, and raise Not_found - so the assert(false) here should be unreachable.

That means that the only place where the "wrong" (in my opinion) exception is raised is in max_binding, which does not raise `Not_found`.

I will reinstate the `assert(false)` and add a comment.

src/batSplay.ml
jabot2: After implementing to_seq and to_seq_from for BatSplay i also used the implementation for BatMap and BatSet.
You can see my implementations in the PR, they look remarkably similar to your implementation in the benchmark.

src/batSplay.ml
jabot2: ... is now implemented in terms of 'a enumeration

jabot2:
Out of curiosity: what is the oldest ocaml version we still need to support with batteries-included?
Can we (e.g) presume that ppx rewriters are available?

src/batSet.ml
```diff
@@ -704,24 +704,26 @@ module Concrete = struct
...
+  let to_seq_from cmp k m () =
+    let rec cons_enum_from k2 m e =
+      match m with
+      | Empty -> e
+      | Node (l, k, r, _) ->
+         if cmp k2 k <= 0
+         then cons_enum_from k2 l (C (k, r, e))
+         else cons_enum_from k2 r e in
+    to_seq_hlp (cons_enum_from k m E)
+
```
gasche, line 726: Now we have two comparable implementations of `to_seq` and `to_seq_from`, this one and the one from #1011. Which one do we want to merge and reuse in the other modules?

My preference would go for merging @jabot2's code (it's always nicer to get code from fresh new contributors than grumpy old ones), but there are aspects of the implementation in #1011 that are better:
- I prefer defining `cons_enum_from` as a toplevel function, close to `cons_iter` (... which suggests that it should be called `cons_iter_from`), because it could be reused for other iterators later.
- currently `to_seq_hlp` returns a `'a BatSeq.node` and the caller takes an additional `()` argument, it would be more efficient if it returned a `'a BatSeq.t` directly. (The problem here is that OCaml assumes that `to_seq_from` is a 4-argument function, which get optimized differently from 3-argument functions.)

(I have a small preference for my naming choice, `seq_of_iter` rather than `to_seq_hlp`, but this is not very important and we could go with your choice.)

gasche:
> Out of curiosity: what is the oldest ocaml version we still need to support with batteries-included?
> Can we (e.g) presume that ppx rewriters are available?

In general I think that the older version that it makes sense supporting is the OCaml version of the current Debian-stable release (or possibly oldstable if it's not too hard). Currently stable has 4.05 and oldstable has 4.02.3. Either of these would be fine.

(On the other hand, using ppx rewriters can introduce a fair bit of maintenance churn/pain, so this is something to decide very carefully.)

src/batSplay.ml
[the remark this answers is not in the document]
gasche: Thanks! (I agree that the exception for `max_binding` was wrong.)

src/batMap.ml
```diff
@@ -53,6 +53,9 @@ module Concrete = struct
 
   let empty = Empty
 
+  let is_empty m =
+    m == Empty
```
gasche, line 57: This should be just `= Empty`. In general we don't want to use physical equality `==` unless when absolutely necessary, because it has a tricky semantics that is less portable and needs more careful code reading.

src/batMap.ml
```diff
@@ -495,7 +577,7 @@ module Concrete = struct
   let filterv f t cmp =
     foldi (fun k a acc -> if f a then add k a cmp acc else acc) t empty
   let filter f t cmp =
-    foldi (fun k a acc -> if f k a then add k a cmp acc else acc) t empty
+    foldi (fun k a acc -> if f k a then acc else remove k cmp acc) t t
```
gasche, line 580: Why did you change the implementation here? If I understand correctly, the previous implementation is better when few keys remain in the result, and the new implementation is better when most keys remain in the result.

Remarks:
- it is odd that `filterv` and `filter` would use opposite algorithms
- we could implement a direct version using `bal` that should be fast in all cases (but then we have already spent a lot of time on this PR so maybe it's okay to keep less-good code if it was already there).

src/batMap.ml
```diff
@@ -806,6 +892,38 @@ module Concrete = struct
...
+  let union_stdlib cmp f m1 m2 =
+    foldi
+      (fun k v m ->
+        update_stdlib k
+          (fun v2opt ->
+            match v2opt with
+            | Some v2 -> f k v v2
+            | None -> Some v) cmp m)
+      m1
+      m2
```
gasche, line 925: This implementation does not have the expected complexity. I think that you could just use `merge` instead, which is more general. (A hand-written implementation of `union` using the same approach as `merge` would also be faster.)

src/batMap.ml
```diff
@@ -1470,6 +1729,20 @@ module PMap = struct (*$< PMap *)
...
+  let to_seq m = Concrete.to_seq m.map
+             
+  let to_seq_from k m =
+    Concrete.to_seq_from m.cmp k m.map
+                      
+  let add_seq s m =
+    { m with map = Concrete.add_seq m.cmp s m.map }
+
+  let union_stdlib f m1 m2 =
+    { m1 with map = Concrete.union_stdlib m1.cmp f m1.map m2.map }
```
gasche, line 1744: This implementation is wrong if `m1.cmp` is not compatible with `m2.map`. See how `heuristic_merge` is implemented.

gasche (commented):
I did a complete review pass over `batMap.ml`.

src/batMap.ml
jabot2: Because if the callback `f` returns true for all bindings then the given map should be returned unmodified.
That behaviour cannot be achieved by starting from `empty` and adding the missing elements...

I can make the implementation of filterv and filter consistent though.

src/batMap.ml
jabot2: ok

src/batMap.ml
jabot2: i used heuristic_merge instead. That also avoids problems when PMaps with incompatible comparisons are merged (see your comment below)

src/batMap.ml
jabot2: avoided by implementing union_stdlib in terms of heuristic_merge instead

src/batSet.ml
gasche: (The implementation was rewritten, taking all these points into account, so it is now perfectly fine.)

src/batSplay.ml
```diff
@@ -769,6 +802,40 @@ struct
...
+    fold
+     (fun k v m ->
+        match find_opt k m with
+        | Some v1 ->
+           (match f k v v1 with
+            | Some vmerged -> add k vmerged m
+            | None -> remove k m)
+        | None -> add k v m)
+      m1
+      m2
```
gasche, line 837: This implementation is not completely satisfying (it does not preserve the knowledge that we build the result in-order), but here the advice to use `merge` instead is not so convincing, because `merge` is okay but not that nice. (It performs an imperative traversal of two `Enum`, instead of working on the nicer, underlying `'a enumeration` structures.) I propose to leave it as-is in the PR, and I maybe submit a subsequent PR to improve `merge` and then define `union` from it.

gasche (approved):
I'm done reviewing the PR and I only had somewhat minor comments left. I believe that this is ready for merging; well done!

What kind of a git-history person are you? Would you like to rebase your commits before merging and merge some of them together, or leave things as-is, or that we squash all of them in a single commit?

src/batMap.ml
gasche: I may rewrite `filter_map` to use ordered join operations instead of this unpleasant `fold`, and define `filter` and `filterv` on top of it. This can wait until the PR is merged.

jabot2:
Considering git: If you prefer a "clean" history, feel free to squash the commits; I personally don't hold a strong opinion on the matter.

PR #1010, opened 2021-01-30
benchmark various implementations of Bench.to_seq discussed in #1006

gasche:
Results on my machine:

    enumeration (89.71 us) is 77.5% faster than
    batseq (399.39 us) which is 3.8% faster than
    too strict (415.29 us) which is probably (alpha=40.06%) same speed as
    simple (416.74 us)
    Saving times to times.flat

This suggests that using an enumeration indeed provides a large
performance advantage over the simple approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,7 @@ BENCH_TARGETS += benchsuite/lines_of.native
 BENCH_TARGETS += benchsuite/bitset.native
 BENCH_TARGETS += benchsuite/bench_map.native
 BENCH_TARGETS += benchsuite/bench_nreplace.native
+BENCH_TARGETS += benchsuite/bench_set_to_seq.native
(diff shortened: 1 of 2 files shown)


gasche:
cc @jabot2

jabot2:
Looks good

UnixJunkie:
Does this mean that in general BatEnum is faster than BatSeq ?
I was hoping that BatSeq is at least as fast as Enum.

gasche:
No, `enumeration` here is an internal datatype that represents a (pervasive) external iterator, and is then used to implement conversion to both Enum and Seq. I am not sure whether Enum is faster or slower than Seq (I think there are cases where the non-allocating nature of the Enum interface makes is faster), but its implementation is certainly more complex and thus harder to trust.

PR #1011, opened 2021-01-30
BatSet: implement to_seq{,_from} using the iteration structure

gasche:
The addition to the testsuite were necessary to catch some potential
bugs in these definitions. (I tried replacing `r` with `l` in the 'then'
branch of cons_iter_from, and without the new tests the testsuite
would still pass.)


gasche:
cc @jabot2

gasche:
This is superseded by the new versions of @jabot2's PR #1006.

PR #1014, opened 2021-02-02
Prefilter dune build

UnixJunkie:
Switching to dune is going to be a long process.
This is a first step.


gasche:
I should probably have said so earlier, but I have a PR in progress to switch to `dune`. This is why I got rid of `rectypes` in finger trees. I would propose that you wait a couple week before putting any more work in the build system, to avoid wasted effort.

UnixJunkie:
You should do it incrementally, there are many changes.
Nobody will want to review the massive PR.
For some time, we will have two build systems living together and that's fine.

gasche:
There is a technical difficulty with splitting the work due to the
prefilter script: ocamlbuild has a rule to produce a preprocessed
`%.ml` from a source `%.mlv`, but Dune expects the input to
have name `%.ml` and does the preprocessing on the fly.

We cannot easily add logic to handle `%.mlv` files to Dune, because as
far as I know Dune does not support parametrized rules.

It would be very hard to support `%.ml` source files in ocamlbuild.

Currently I manually create symbolic links from `foo.ml` to `foo.mlv`
to test the Dune build, this works well but it breaks the ocamlbuild
build. I will see about scripting that if we want to support both
systems during a transition period.

UnixJunkie:
I strongly dislike the .mlv and .mliv files.
They should all be renamed to .ml and .mli and the build system should use the -pp option correctly.

UnixJunkie:
This also should come in a separate PR:
```
##V<4.4##external unsafe_blit :  t -> int -> t -> int -> int -> unit = "caml_blit_string" "noalloc"
##V>=4.4##external unsafe_blit :  t -> int -> t -> int -> int -> unit = "caml_blit_bytes" [@@noalloc]
```
The whole line doesn't need to be duplicated: only the last keyword should be made conditional.

UnixJunkie:
partially superseded (build/dune is missing) by https://github.com/ocaml-batteries-team/batteries-included/pull/1016

PR #1016, opened 2021-02-03
Dune preliminaries: fix Dune warnings over the Batteries codebase

gasche:
@UnixJunkie suggested in #1014 to split the migration-to-dune work in several smaller PRs. This is the first PR. There are no dune files in this PR, but only changes that needs to be fixed before a dune build can work:
- `opam` is renamed `batteries.opam`
- some fixes in `.gitignore`
- a *lot* of warnings activated-by-default-in-dune are fixed

(*one* unused-variable warning actually caught a small bug, that is fixed in this PR in its own commit for visibility).


UnixJunkie:
This PR is too big.
It would be easier (and very fast) to review separately: unused variables, unused rec keyword, ocamldoc problems, etc.

UnixJunkie:
This also should come in a separate PR:

##V<4.4##external unsafe_blit :  t -> int -> t -> int -> int -> unit = "caml_blit_string" "noalloc"
##V>=4.4##external unsafe_blit :  t -> int -> t -> int -> int -> unit = "caml_blit_bytes" [@@noalloc]

The whole line doesn't need to be duplicated: only the last keyword should be made conditional.

gasche:
> This PR is too big.

Have you tried reviewing commit-per-commit?

I guess the issue may be with the "fix a lot of warnings" commit. If you want, I could split then with one commit per file, but splitting them by kind of warnings would be too much work. (I think they are okay to review as-is by skimming, as most of the changes are visually obvious.)



> This also should come in a separate PR:
> 
> ##V<4.4##external unsafe_blit : t -> int -> t -> int -> int -> unit = "caml_blit_string" "noalloc"
> ##V>=4.4##external unsafe_blit : t -> int -> t -> int -> int -> unit = "caml_blit_bytes" [@@noalloc]
> 
> The whole line doesn't need to be duplicated: only the last keyword should be made conditional.

The fact that there are two lines was already there in the code, it does not come from the PR. The reason is that the primitive names are different. I use "noalloc" in one situation and [@@noalloc] in the other because the "noalloc" form was deprecated in OCaml 4.03.

UnixJunkie:
> > This PR is too big.
> 
> Have you tried reviewing commit-per-commit?

No, I use github's graphical diff to review PRs.

> I guess the issue may be with the "fix a lot of warnings" commit. If you want, I could split then with one commit per file, but splitting them by kind of warnings would be too much work. (I think they are okay to review as-is by skimming, as most of the changes are visually obvious.)

Do you mind if I break your branch into several PRs?

Could you add to ocamlbuild all the warnings dune uses, so that both build systems get the same
warning as errors?

> > This also should come in a separate PR:
> > ##V<4.4##external unsafe_blit : t -> int -> t -> int -> int -> unit = "caml_blit_string" "noalloc"
> > ##V>=4.4##external unsafe_blit : t -> int -> t -> int -> int -> unit = "caml_blit_bytes" [@@noalloc]
> > The whole line doesn't need to be duplicated: only the last keyword should be made conditional.

Sorry, I did not see that. But to my defense, I would probably have seen it in a smaller, dedicated PR...

> The fact that there are two lines was already there in the code, it does not come from the PR. The reason is that the primitive names are different. I use "noalloc" in one situation and [@@noalloc] in the other because the "noalloc" form was deprecated in OCaml 4.03.

gasche:
> No, I use github's graphical diff to review PRs.

If you want to review per-commit:
- click on the list of commits
- open each commit in a tab
- for each commit, the same graphical diff interface is available as for the whole PR
- at any point you can send a partial review, mentioning the last commit reviewed (or where you stopped in the last commit reviewed), to easily start again later

> Could you add to ocamlbuild all the warnings dune uses, so that both build systems get the same
warning as errors?

Sounds reasonable.

gasche:
> Do you mind if I break your branch into several PRs?

No, I certainly don't mind.

UnixJunkie:
I will try to do it this week.

UnixJunkie:
I will continue merging later.
One yet unmerged commit is a (probably undetected) bug correction: the function was using Std.compare while the user was giving a specialized comparator function.

UnixJunkie:
All has been merged now.
Just one remark: why in batFloat.mli we have two times 'include BatNumber.Bounded [...]'. Is this really necessary?

UnixJunkie:
Also, do we just miss a dune file now?

PR #1023, opened 2021-02-10
rm trailing semicolons (unnecessary)

UnixJunkie:
all changes were extracted from
https://github.com/gasche/batteries-included.git#dune-preliminary-warnings


gasche:
Note: they are not only unnecessary, they result in a warning because the documentation string that follows is not correctly attached to the signature item.

PR #1027, opened 2021-02-19
add new Either module if OCaml >= 4.12

UnixJunkie:


UnixJunkie:
Maybe, we should copy-paste the .ml file from the stdlib directly, and not make this new module conditional on the ocaml version.

UnixJunkie:
@gasche what do you think?

gasche:
Indeed, you should not use a `V>=4.12` conditional for this module, as it works with older OCaml versions. (Same for `List.partition_map`.)

UnixJunkie:
ok, I'll change that.

UnixJunkie:
Shouldn't we have added a constraint to specify that Eiter.t = BatEither.t, somehow?
For ocaml >= 4.12.

UnixJunkie:
I'll send a PR, we have the same constrain in BatSeq.

PR #1028, opened 2021-02-19
BatList.partition_map: tail rec. version of a new stdlib function

UnixJunkie:


gasche (approved):
I believe that the code is correct.

(Note: stdlib version is already tail-recursive, the difference here is that you use the `Accum` machinery of Batteries to write it in destination-passing style, which should be more efficient than the stdlib's reverse-at-the-end approach.)

UnixJunkie:
I know that on very long lists, batteries way is faster.
Thanks for the review.

gasche:
Yes, you might be interested in the detailed benchmarks in my recent TRMC paper: [Tail Modulo Cons](http://gallium.inria.fr/~scherer/doc/tmc-jfla-2021.pdf).

UnixJunkie:
Interesting paper, very OCaml-specific though, where is it published?

gasche:
It was accepted at [JFLA'21](http://jfla.inria.fr/jfla2021.html), a French-speaking conference on functional programming. (It is small in scale but a rather interesting event.)

PR #1033, opened 2021-03-01
DynArray.remove_at alias

fccm:


UnixJunkie:
great, but I will only merge after the current release, which is already underway

UnixJunkie:
Hi Florent, please describe your change in the ChangeLog and I will merge this.

PR #1034, opened 2021-03-01
add Buffer.add_substring'

### LINKED ISSUE #934: BatBuffer: No way to add BatSubstring.t
Not sure if there's some reason behind this, but it's seemingly impossible to add a BatSubstring to a BatBuffer right now. Further, BatBuffer.add_substring is named confusingly because of that. Am I missing something here?

OhadRau:
This function allows adding a BatSubstring.t to a BatBuffer.t, and is otherwise analogous to BatBuffer.add_substring. It resides in its own module, BatBufferHelper, in order to avoid a cyclic dependency error, as BatSubstring indirectly depends on BatBuffer. In batteries.mlv, we combine the two modules to export as Buffer.

Fixes #934 

I'm not sure this interface is ideal still for a number of reasons, so I would appreciate some additional feedback regarding the following:
* Is `Buffer.add_substring'` the best name for this function? I suggested a few alternatives in #934
* Is the way I avoid the cyclic dependency error ok code style for batteries? I played around with some other ways of doing this and found this one to be most desirable, but happy to change around if this feels "icky." The main downside here is that ocamldocs won't show `BatBuffer.add_substring'` and users will need to search for it in the `BatBufferHelper` module


UnixJunkie:
Nice hack to avoid the circular dependency.
The prime character is missing in the ocamldoc.
Do other maintainers have a better name to suggest?

UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/934

UnixJunkie:
My suggestion to avoid the hack is to put the function inside of BatBuffer, if that's possible.
Others maintainers might have other suggestions though.

gasche:
I think the cycle-avoidance approach currently in the PR is not going to work, because many users do not `open Batteries`, they use BatFoo directly.

One way to solve those cycles -- but it can be heavyweight -- is to have a BatInnerSubstring module that has very few dependencies (in particular, no Bat* dependencies), that BatBuffer can depend on, and is included in BatSubstring.

UnixJunkie:
If the new function does not show up in the ocamldoc, this is not acceptable also (users will never know it exists).

OhadRau:
@gasche Thanks for the suggestion, seems like a reasonable way to go about it and should take care of the documentation issue :) I just pushed the fix (though apologies, I messed up my git history a little bit there; let me know if you want me to split making BatInnerSubstring into its own commit)

UnixJunkie:
Yes, can you get rid of the changes in build/intro.txt.
Also, you need to update the ChangeLog before we can merge this.
In principles, there should be a unit test also.

UnixJunkie:
Don't you need to add a new file like src/batInnerSubstring.ml ?

OhadRau:
> 
> 
> Don't you need to add a new file like src/batInnerSubstring.ml ?

Should be there already: https://github.com/ocaml-batteries-team/batteries-included/pull/1034/files#diff-159a2fa030115fb22a8ad6e24b26627701e01e58212eb7326a51418f72f1e1e5

Just deleted the build/intro.txt change, updated ChangeLog & added a unit test. Lmk if there's anything else I should change

UnixJunkie:
I don't think, because of its name, that BatInnerSubstring was ever meant to become a public module.
I might send another proposition.

UnixJunkie:
Let's keep it simple?
https://github.com/ocaml-batteries-team/batteries-included/pull/1035

UnixJunkie:
@gasche You can merge this PR if you like.
In which case, I will just close my half-joking competing PR.

gasche:
My time to contribute to Batteries is limited these days, and I sort of burned it out on the Dune migration work (yet to be finished!). I think that the PR in its current state is not completely satisfying (I regret that the change in module structure is exposed in `BatSubstring.mli`), and I am not sure what is the best approach.
Another minor issue is the unpleasant name `add_susbtring'`, which I am not sure how to avoid (so maybe there is nothing we can do against it).

I wonder if we should consider simply exposing the definition of `BatSubstring` as triple `(string * int * int)`. Because this type is structural, it could be the input type of `BatBuffer.add_substring'` without introducing any dependency on BatSubstring. But this is also not fully satisfying.

These things are very hard...

UnixJunkie:
@gasche Maybe consider my dumb proposal then: https://github.com/ocaml-batteries-team/batteries-included/pull/1035/files

PR #1037, opened 2021-04-19
Exceptionless extended with reduce, min and max functions

nicoTolly:
So I added a few functions to the ExceptionLess modules (with the obvious semantics). Tell me if there is anything wrong with the implementation.
I also took the liberty to generalize a bit the standard min and max functions by adding a default cmp argument, so they are consistent with the current definition of min_max function. Tell me if you think this change is unneeded or unwanted (maybe an optional argument has some runtime cost ? I actually have no idea). I can also put them in a separate PR.

I also tried to add some unit test but failed to have them compiling. I added this in the ExceptionLess module : 
```
  (*$T max
    max [] = None
    max [1] = Some 1
    max [1; 1] = Some 1
    max [1; -2; 3; 4; 5; 60; 7; 8] = Some 60
  *)
```
but the compiler complained with : 
```
File "src/batList.mlv", line 1635, characters 14-20:
Error: This expression has type 'weak89 option
       but an expression was expected of type int
```

Could you tell me how to fix that ?
Thanks

### DIFF
--- a/src/batList.mliv
+++ b/src/batList.mliv
@@ -1073,6 +1073,29 @@ module Exceptionless : sig
   val last : 'a list -> 'a option
     (** [last l] returns either [Some x] where [x] is the last element of the list, or [None] if
         the list is empty. This function takes linear time. *)
+
+  val reduce : ('a -> 'a -> 'a) -> 'a list -> 'a option
+  (** [reduce f h::t] is [Some (fold_left f h t)]
+      and [reduce f []] is None. *)
+
+  val min_max : ?cmp:('a -> 'a -> int) -> 'a list -> ('a * 'a) option
+  (** [min_max l] returns either Some(s, l)
+      where s and l are respectively the smallest and biggest element of [l] as judged by
+      [Pervasives.compare] (by default) or None if [l] is empty.
+	  You can provide another comparison function via the optional [cmp] parameter.
+  *)
+
+  val max : ?cmp:('a -> 'a -> int) -> 'a list -> 'a option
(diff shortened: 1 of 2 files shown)


UnixJunkie:
Thanks a lot for this contribution and welcome to batteries!

I will look into it soon, if no other maintainer does.
For the type error, you need to tell the compiler the type of the empty list or constrain the type of the option
in the empty list case of your tests.
If you cannot, don't bother and I will fix it.

nicoTolly:
> For the type error, you need to tell the compiler the type of the empty list or constrain the type of the option
> in the empty list case of your tests.
> If you cannot, don't bother and I will fix it.

Sorry, this is not the point, I should have been more explicit. The problem is that the compiler apparently can't tell that there are two different max function : the standard one with type 'a list -> 'a and the one in ExceptionLess module with type 'a list -> 'a option.

Thus, even if I do this : 
```ocaml
	(*$T max
	 max [1] = (Some 1 : int option)
	 *)
```
I trigger this error : 
```
File "src/batList.mlv", line 1617, characters 12-33:
Error: This expression has type int option
       but an expression was expected of type int
```

UnixJunkie:
I also don't know how to fix the testing problem you describe.
But your PR has been merged.

UnixJunkie:
I just called you nicoTolly in the ChangeLog file; if you want your real name in it, please tell us.

nicoTolly:
@UnixJunkie Great ! Please tag me with my real name, Nicolas Tollenaere :) Thanks !

UnixJunkie:
done!

PR #1044, opened 2021-09-02
Fix multicore support

kit-ty-kate:
https://github.com/ocaml-multicore/ocaml-multicore/pull/514 broke the multicore support of batteries

### DIFF
--- a/build/prefilter.ml
+++ b/build/prefilter.ml
@@ -37,6 +37,12 @@ let print_loc = function
        stale := false;
      end
 
+let has_domains ~extra =
+  let extra = String.split_on_char '+' extra in
+  List.mem "domains" extra ||
+  List.mem "multicore" extra ||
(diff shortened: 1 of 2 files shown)


build/prefilter.ml
```diff
@@ -37,6 +37,12 @@ let print_loc = function
        stale := false;
      end
 
+let has_domains ~extra =
+  let extra = String.split_on_char '+' extra in
+  List.mem "domains" extra ||
+  List.mem "multicore" extra ||
+  List.mem "effects" extra
```
gasche, line 44: This may have to be refined again in the future, given that `effects` may add extra stdlib logic in addition to just `domains`. But I'm okay with going for this for now, and refining as it comes in the future.

gasche (approved):

UnixJunkie:
Thanks for those fixes for the future.

Issue #1045, opened 2021-09-24
OCaml 4.13 support


kit-ty-kate:
OCaml 4.13 is being released as we speak today and batteries, despite being "compatible" (not on opam though) does not pass its tests as the new functions from 4.13 are not exported.

UnixJunkie:
I will try to work on it this week.

UnixJunkie:
much work remains...
```
File "src/batteries_compattest.ml", line 13, characters 18-23:
13 |   module Bytes = (Bytes : module type of Legacy.Bytes)
                       ^^^^^
Error: Signature mismatch:
       ...
       The value `fold_left' is required but not provided
       File "bytes.mli", line 185, characters 0-55: Expected declaration
       The value `fold_right' is required but not provided
       File "bytes.mli", line 191, characters 0-56: Expected declaration
       The value `for_all' is required but not provided
       File "bytes.mli", line 197, characters 0-45: Expected declaration
       The value `exists' is required but not provided
       File "bytes.mli", line 201, characters 0-44: Expected declaration
       The value `starts_with' is required but not provided
       File "bytes.mli", lines 347-348, characters 0-77: Expected declaration
       The value `ends_with' is required but not provided
       File "bytes.mli", lines 354-355, characters 0-77: Expected declaration
       The value `split_on_char' is required but not provided
       File "bytes.mli", line 490, characters 0-46: Expected declaration
```
and probably in many more modules too.

UnixJunkie:
I'll try to contribute some more tomorrow:
```
File "src/batteries_compattest.ml", line 31, characters 18-23:
31 |   module Int32 = (Int32: module type of Legacy.Int32)
                       ^^^^^
Error: Signature mismatch:
       ...
       The value `min' is required but not provided
       File "int32.mli", line 218, characters 0-20: Expected declaration
       The value `max' is required but not provided
       File "int32.mli", line 223, characters 0-20: Expected declaration
```

UnixJunkie:
Many new things in the String module form the stdlib. Will take some time to fix those in the compatibility test...
```
       The value `empty' is required but not provided
       File "string.mli", line 98, characters 0-18: Expected declaration
       The value `of_bytes' is required but not provided
       File "string.mli", line 104, characters 0-30: Expected declaration
       The value `to_bytes' is required but not provided
       File "string.mli", line 111, characters 0-30: Expected declaration
       The value `cat' is required but not provided
[...]

PR #1046, opened 2021-09-27
BatArray.fold_left_map

UnixJunkie:
introduced in stdlib from OCaml-4.13.0


UnixJunkie:

PR #1048, opened 2021-09-27
BatArray.find_map

UnixJunkie:
introduced in stdlib since OCaml-4.13.0

### DIFF
--- a/src/batArray.mliv
+++ b/src/batArray.mliv
@@ -479,6 +479,11 @@ val find : ('a -> bool) -> 'a array -> 'a
     @raise Not_found if there is no value that satisfies [p] in
     the array [a]. *)
 
+val find_map : ('a -> 'b option) -> 'a array -> 'b option
+(** [find_map f a] applies [f] to the elements of [a] in order, and returns the
+    first result of the form [Some v], or [None] if none exist.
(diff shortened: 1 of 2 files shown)


UnixJunkie:

gasche:
@UnixJunkie did you write this implementation yourself, or did you copy it from the stdlib? If you copied the stdlib implementation, I believe that this should be marked clearly (copying code without referencing the author is, in fact, against the law). One way to mark this is to credit the corresponding commit to the original author of the code: `git commit --author "..."` (in this case, I think it is Nicolás Ojeda Bär). When you make changes to someone else's implementation, you can use a `Co-authored-by: ...` line in the commit message (I would use the original author in the `--author` line, and then `Co-authored-by: <myself>`).

gasche:
One other reason to be explicit when the code is copied from the stdlib is that it helps reviewing, other people can focus on the commits that provide new or modified implementations.

UnixJunkie:
The implementation and ocamldoc was copied from the stdlib.
The tests are mine, same for some boilerplate code.
I am OK to put a one-liner comment to say that it comes from the stdlib of 4.13.0 and also
mention it in the ChangeLog.
However, having to go find the author of each new functionality would be quite annoying.

PR #1049, opened 2021-09-27
BatArray.combine

UnixJunkie:
introduced in stdlib since OCaml-4.13.0


UnixJunkie:

PR #1053, opened 2021-10-21
Compat 4.13.1

UnixJunkie:


UnixJunkie:
Many fixes for compatibility with the latest stdlib.
I would like if another pair of eyes can look at those.
I should merge tomorrow.

PR #1062, opened 2021-12-14
Release 3.4.0

kit-ty-kate:


UnixJunkie:
Ok, that's part of the release process.
The ChangeLog is probably missing several PRs that were merged.
I would have to scheme through the git logs to add the missing pieces.
Also, the most important part is testing 'make test' works without any error on several OCaml versions (of course the one we want to support, but also older ones we are still supporting).

kit-ty-kate:
for `make test` I’ve only tried on macOS/arm64 with OCaml 4.13, but why not just let opam-repo-ci handle that when publishing in opam? I’m not sure why the tests isn’t listed in the opam file but it should be good enough. Also we have linux x86_32, arm32, arm64, x86_64, s390x, ppc64 and macOS x86_64, that should cover most of the platforms

UnixJunkie:
If you use the opam CI for this purpose, then better tag with a temporary release tag.
Once all the CI is OK, then use the final release tag; this will prevent you from having to bump
the release tag (in case you have to fix some things because of failing tests after having burned/used the release tag).

kit-ty-kate:
Sure, done in https://github.com/ocaml/opam-repository/pull/20225

UnixJunkie:
@kit-ty-kate you could send a PR with your name added to the list of maintainers (in the opam file); you have been contributing to batteries several years IIRC

PR #1065, opened 2021-12-15
Add covariances to types of maps and sets

eponier:
I added covariance annotations to types of maps and sets. I also tried to add some tests that check that this allows to pass the value restriction and get proper polymorphism instead of weak polymorphism.

Maybe some other types in other files can be annotated like that, I had a look but did not do an extensive search.

### DIFF
--- a/src/batMap.mliv
+++ b/src/batMap.mliv
@@ -534,7 +534,7 @@ module String : S with type key = string
     function, it is recommended to use the functorized maps provided
     by {!Make}.  *)
 
-type ('a, 'b) t
+type (+'a, +'b) t
 
 val empty : ('a, 'b) t
 (** The empty map, using [compare] as key comparison function. *)
@@ -973,7 +973,7 @@ module PMap : sig
       functor for additional safety.
   *)
 
-  type ('a, 'b) t
+  type ('a, +'b) t
 
   val empty : ('a, 'b) t
   (** The empty map, using [compare] as key comparison function. *)
--- a/src/batSet.mli
+++ b/src/batSet.mli
@@ -516,7 +516,7 @@ module String : S with type elt = string
     @author David Rajchenbach-Teller
 *)
 
-type 'a t
+type +'a t
 (** The type of sets. *)
 
 include BatEnum.Enumerable with type 'a enumerable = 'a t


UnixJunkie:
Are those annotations supported by all OCaml versions?
Currently, batteries declares in the opam file that it is compatible with OCaml >= 4.00.0.

eponier:
Apparently, it was added in 3.01 (see https://github.com/ocaml/ocaml/blob/98b277c2d55eb820e88a5447cc5cb2e9d6bdfacd/Changes#L10922). I didn't think it was this old. But in fact, it is already used in multiple files of batteries (including BatMap.mlv), so I think we are good on this aspect.

eponier:
I am more worried by the tests I have added, because I didn't know qtest before. I am not sure that I followed the spirit of the tool, perhaps it is better to remove them.

UnixJunkie:
Honestly, I don't know what those type annotations mean.
So, I have no idea on if and how they should be tested (isn't compilation enough to check type-related things?).
```
type (+'a,-'b,'c) t (covariant in 'a, contravariant in 'b, invariant in 'c)
```
doesn't mean anything to me (I am not a CS researcher; @gasche should know more).
If those type annotations are useful, because they make the type of the data-structure more generic, then that's good and such
annotations should be added to all data-structures where possible.
Also, so they are merged in the next minor release, it must not break things in existing users' code.

eponier:
If you have a few minutes, everything is explained in the [OCaml doc](https://ocaml.org/manual/polymorphism.html#s:weak-polymorphism). In a few words, the fact that a type is used in covariant position means that you can replace it by a larger type. But it also has an impact on how the type-checker infers polymorphism: if it knows that a type variable occurs in covariant position, then it can give a more generic type. When you have a concrete type, OCaml infers the variance automatically, but when you have an abstract type, you have to specify the expected variance. As usual, the type-checker checks at the instantiation point that the concrete type respects the given signature, and this includes the variance. (The idea of the test was not to verify that the type variable was covariant, this is the role of the type-checker indeed, but to make use of this variance annotation so that someone trying to change/remove it can see an impact on the tests.)

This makes the types less generic, actually, meaning that it's easier to user them, but potentially harder to produce them. If we consider that everyone uses the functor provided by batteries, it shouldn't be a problem. If someone decided to reimplement the functor using a type where the type variable is not covariant, then the module will not abide by the signature anymore. I don't known if some people do that.

eponier:
To be complete, my initial motivation was to be able to have better polymorphism.
For instance, before the change
```
let empty = (fun x -> x) BatMap.empty;;
val empty : ('_weak1, '_weak2) BatMap.t = <abstr>
BatMap.add 1 2 empty;;
- : (int, int) BatMap.t = <abstr>
empty;;
- : (int, int) BatMap.t = <abstr>
```
This is weak polyphormism, meaning that only one type is allowed, initially unknown. As soon as some operation constrains the type, it is fixed.

After the change
```
let empty = (fun x -> x) BatMap.empty;;
val empty : ('a, 'b) BatMap.t = <abstr>
BatMap.add 1 2 empty;;
- : (int, int) BatMap.t = <abstr>
empty;;
- : ('a, 'b) BatMap.t = <abstr>
```
This is true polymorphism, `empty` has a generic type.

UnixJunkie:
Do you want to update your PR?
Do you want to change the tests / remove them?

eponier:
Yes, I removed the tests. I don't think that they are necessary and they were a bit artificial.

UnixJunkie:
thanks a lot for your contribution

PR #1066, opened 2021-12-15
opam file: Fix constraint (uses QCheck_ounit) and add missing conflict with some compiler variants

kit-ty-kate:
Upstreams changes made in https://github.com/ocaml/opam-repository/pull/20225

### DIFF
--- a/batteries.opam
+++ b/batteries.opam
@@ -23,9 +23,13 @@ depends: [
   "ocamlfind" {build & >= "1.5.3"}
   "ocamlbuild" {build}
   "qtest" {with-test & >= "2.5"}
-  "qcheck" {with-test & >= "0.6" & < "0.14"}
+  "qcheck" {with-test & >= "0.9" & < "0.14"}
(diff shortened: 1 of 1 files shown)


batteries.opam
```diff
@@ -23,10 +23,14 @@ depends: [
...
   "benchmark" {with-test & >= "1.6"}
   "num"
 ]
+conflicts: [
+  "base-effects"
+  "ocaml-option-no-flat-float-array"
+]
 # url {
 #   src: "https://github.com/ocaml-batteries-team/batteries-included/archive/vXXX.tar.gz"
 #   checksum: "md5=YYY"
```
kit-ty-kate, line 36:
```suggestion
```
Also this is a bit unnecessary and surprising for users looking for examples. Using `opam publish` is much easier to do this job rather than doing it by hand.

UnixJunkie: you can change md5 to something else, if that's your concern

PR #1069, opened 2022-01-19
BatHashtbl.for_all

UnixJunkie:


UnixJunkie:
If somebody can eyeball through this PR, this would be nice.
I'll update the changelog if accepted.

PR #1071, opened 2022-01-19
BatFilename.with_temp_file

UnixJunkie:


src/batFilename.mlv
```diff
@@ -62,3 +62,16 @@ let split_extension s =
   ("/foo/.rc", "")    (split_extension "/foo/.rc")
   ("", "")            (split_extension "")
 *)
+
+let with_temp_file ?temp_dir prfx sfx f =
+  let tmp_fn = Filename.temp_file ?temp_dir prfx sfx in
+  let res = f tmp_fn in
+  Sys.remove tmp_fn;
+  res
```
rixed, line 70: Why not use `finally` here to close the file regardless of [f] raising an exception?

UnixJunkie: ok, so we will remove the created file even if f crashed

PR #1075, opened 2022-01-25
Mark as compatible with OCaml 4.14

kit-ty-kate:


kit-ty-kate:
All green in the opam-repository draft PR and ready to be tagged

Issue #1078, opened 2022-02-23
Expose `Hashtbl.stats`


sim642:

UnixJunkie:
the stdlib exposes it, so I guess it is OK to do the same
```ocaml
val stats : ('a, 'b) t -> statistics
(** [Hashtbl.stats tbl] returns statistics about the table [tbl]:
   number of buckets, size of the biggest bucket, distribution of
   buckets by size.
   @since 4.00.0 *)
```

UnixJunkie:
@sim642 if you can review the related PR and test it, that would be nice.
Batteries is always looking for people to help with the maintenance.

UnixJunkie:
done

PR #1079, opened 2022-02-24
BatHashtbl.stats exposed

UnixJunkie:
ocamldoc imported from the stdlib


UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/1078

sim642 (approved):
Works!

I'm just wondering if maybe the `statistics` type should also be re-exported? Might make sense if `Batteries` has been opened.

UnixJunkie:
Ok, I will also add the type declaration in the interface file.

PR #1080, opened 2022-03-30
Improved a bit the error when Random.choice is called on an empty enum

nicoTolly:
Hi,

Until now this piece of code : 
```ocaml
 Random.choice @@ Enum.empty ()
```
returned an Exception Invalid_argument "Random.int" which is confusing - the error happened in function Random.int to which we pass the invalid argument 0.
I changed it to raise an exception Empty instead - which matches what happens in similar cases in most other modules. I hope the error message will then be a bit more tractable.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,10 @@ Changelog
 
 ## v3.5.1 (minor release)
 
+- improved BatRandom.choice error a bit when passing an empty enum
+#1080
+ (Nicolas Tollenaere)
+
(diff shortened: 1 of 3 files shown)


UnixJunkie:
can you use an if-then-else to make the code more explicit?

UnixJunkie:
you should add an entry to the ChangeLog also

nicoTolly:
Is it better ?

UnixJunkie:
ok, I'll do some syntactic changes just after

PR #1082, opened 2022-06-13
Support new OCaml 5 unix primitive names

dra27:
The compiler is considering ensuring all symbols in the Unix library are prefixed `caml_` for 5.0 (https://github.com/ocaml/ocaml/pull/10926).

This affects batteries only for the `unix_link` primitive which is updated here. I couldn't see that there was a way to specify a range of versions for `##V` markers, so apologies for the slightly convoluted way of selecting between 3 different `external` declarations!

### DIFF
--- a/src/batUnix.mlv
+++ b/src/batUnix.mlv
@@ -23,7 +23,8 @@
 include Unix
 
 ##V<4.8##external link : string -> string -> unit = "unix_link"
-##V>=4.8##external link : ?follow:bool -> string -> string -> unit = "unix_link"
+##V>=4.8####V<5.0##external link : ?follow:bool -> string -> string -> unit = "unix_link"
+##V>=5.0##external link : ?follow:bool -> string -> string -> unit = "caml_unix_link"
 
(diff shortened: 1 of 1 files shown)


UnixJunkie:
I think we can specify a range; I'll find back.

UnixJunkie:
after checking in https://github.com/ocaml-batteries-team/batteries-included/blob/master/build/prefilter.ml; apparently not

UnixJunkie:
but, there is a use example in src/batteries_compattest.mlv:
```ocaml
##V>=4.7####V<4.14##` module Seq = (Seq : module type of Legacy.Seq)
```

UnixJunkie:
If you can use this kind of syntax and your tests work; I'll merge your proposed change

dra27:
Ah, thanks - I didn't spot that! Updated to use it, and I checked that it's building correctly on 4.12 still and pre-processing correctly on 5.0 (there are other fixes needed for OCaml 5 support - I'm only dealing with the bit I've broken here 🙂)

Issue #1081, opened 2022-05-23
OCaml 5 support


sim642:
Currently the upper bound on ocaml forbids version 5.0:
https://github.com/ocaml-batteries-team/batteries-included/blob/77bd2aace3e16499ec71bf544270e169656c3c23/batteries.opam#L22

Although OCaml 5 isn't yet final, many libraries have been made compatible already. Would be nice if batteries could also do the same.

Maybe it would just be a version bump, maybe it requires more changes. Either way, an issue to track that progress doesn't hurt.

### TODO
- [ ] `ocamlfind: Package bigarray not found` — bigarray is directly inside `Stdlib` since OCaml 4.07, but OCaml 5 removes the package definition (https://github.com/ocaml/ocaml/pull/10896).
- [ ] `Error: Unbound module Pervasives` — OCaml 5 removes the deprecated `Pervasives` in favor of `Stdlib` (https://github.com/ocaml/ocaml/pull/10896).
- [ ] OCaml 5 removes `Stream` and `Genlex` (https://github.com/ocaml/ocaml/pull/10896) so `BatStream` and `BatGenlex` shouldn't be available on OCaml 5 either?

UnixJunkie:
Yeah, doesn't hurt to track our compatibility.

UnixJunkie:
If you can do 'make -k' in your batteries checkout with an OCaml-5 compiler and post the output here, it might tell us how much work is needed.

sim642:
Gave that a shot, after installing all non-ocaml dependencies (which luckily are available on OCaml 5).
First hit the issue about bigarray package not being found, which was easy to work around, but `Pervasives` being removed probably requires a lot more changes (and thinking how to bake it backwards-compatible in batteries). Added these as tasks to the initial issue description.

UnixJunkie:
Can you copy-paste the whole compilation log?
Or link a gist with it?

sim642:
There's not much, out of the box `make -k` outputs:
```
Build mode: shared
ocamlbuild -no-links -use-ocamlfind src/batteries.cma toplevel/batteriesHelp.cmo src/batteriesThread.cma META src/batteries.cmxs src/batteriesThread.cmxs
Finished, 0 targets (0 cached) in 00:00:00.
+ ocamlfind ocamldep -package bigarray,num,str -package bytes -modules src/batOpaqueInnerSys.ml > src/batOpaqueInnerSys.ml.depends
[...]

PR #1085, opened 2022-07-21
Make it compile on OCaml 5

sim642:
Related to issue #1081.

I've brutally massacred our boy to make it compile (and pass tests) on OCaml 5:

1. Use stdlib-shims and campl-streams for some backwards compatibility. They have a `ocaml >= 4.02.3` lower bound instead of Batteries' current `ocaml >= 4.00.0`, but that should be reasonable enough if a new major version is to be supported.
2. Naively replace `Pervasives` with `Stdlib`. Of course this changes Batteries' interface, but thanks to stdlib-shims that should work for the entire version range. This replacement makes up most of the diff.
3. Delete bunch of deprecated functions that OCaml 5 removed. I suppose some more reasonable backwards compatibility could be provided for OCaml 4, but if there's going to be a new major Batteries version, then it might be easier to just bite the bullet and make lots of breaking changes at once.
4. Disable a few tests that otherwise fail, probably due to changes to `Random`.

**As-is, this definitely should not be merged.** But it provides a tentative overview of the changes that would be necessary, but probably is broken on OCaml 4 right now.


UnixJunkie:
Thanks a lot for the trial.
I will have a look at your thing.
Of course, the PR is huge, so reviewing will take some time.

UnixJunkie:
Maybe I'll create an ocaml5 branch and try to merge your work in there, little by little.

Sudha247:
Thanks for your work @sim642! I'd imagine this would take time for reviewing as @UnixJunkie mentioned above, owing to the size of its changes. Meanwhile, if this branch works with `5.0.0~alpha1` switch, feel free to make a patch to add it to https://github.com/kit-ty-kate/opam-alpha-repository/tree/master/packages

UnixJunkie:
If the PR was broken in to several, each dealing with a distinct aspect, this would significantly accelerate the review process.
For example: one PR fixes all compilation errors with the same error message.

UnixJunkie:
I am starting to look at this right now.
Not sure how long it will take me.
I will do several partial merge I guess (if I can group things into related concerns).

UnixJunkie:
the num library can still be installed; even with ocaml-5:
```bash
opam pin add num https://github.com/ocaml/num.git
```

sim642:
> I am starting to look at this right now.
> Not sure how long it will take me.
> I will do several partial merge I guess (if I can group things into related concerns).

Thanks! Looking at this commit by commit is probably the easiest anyway because most of the diff is the `Pervasives`→`Stdlib` rename, which makes other changes hard to notice.

If I remember correctly, the commits are in an order to fix one build problem after another, as they come up. The dirtiest commit is https://github.com/ocaml-batteries-team/batteries-included/pull/1085/commits/1b0c17734f3ec966baf353d5e51744555b948a23 because there I just deleted now-removed deprecated functions, but didn't add all the newer non-deprecated alternatives that they have in `Stdlib`, which should eventually be done.

UnixJunkie:
superseded by https://github.com/ocaml-batteries-team/batteries-included/pull/1086

UnixJunkie:
@sim642 I think I just used your proposed changes for BatFormat.
You are welcome to review the PR I just mentioned.
The diff should be way smaller compared to what you initially proposed.

Issue #906, opened 2019-08-19
rm the _oasis file


UnixJunkie:
anyone against this?
Then simplify release.md.

UnixJunkie:
apparently the opam build uses the setup.ml file generated by oasis.
I don't know why, but I feel that if we could get rid of it, that would be nice.
I.e. we cut permanently the dependency to oasis.

UnixJunkie:
If some maintainers are against this, please raise your hand.

gasche:
I think it's perfectly fine to remove the `_oasis` if we can have a working build/install process without it. Relatedly, one should probably think of switching to dune, which would provide some similar services to Oasis (in particular, an easy way to build the docs).

UnixJunkie:
The current _oasis file is just a wrapper calling our Makefile... So, it seems completely superfluous.
But removing it should be done with care, and we should grep for oasis and _oasis every where...

UnixJunkie:
I would also love to switch to dune.
It is tracked in another issue: https://github.com/ocaml-batteries-team/batteries-included/issues/738

UnixJunkie:
also look for occurences of setup.ml; a file which is generated by oasis

UnixJunkie:
it also generate the META file from the META.in file; apparently just to update the version tag

UnixJunkie:
compiles on 4.02.3

UnixJunkie:
(running the unit test suite requires ocaml 4.03.0 at least because of qcheck-ounit, I believe)

UnixJunkie:
compiles and all tests are passing w/ ocaml 4.13.1

UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/1081

UnixJunkie:
I think the generation of the META file should be delegated to the myocamlbuild.ml script.

PR #1086, opened 2022-08-24
Ocaml5 wip 01

UnixJunkie:


UnixJunkie:
Dear maintainers, note that this just fixes compilation for OCaml 5.
I did not run the unit test suite yet.
Reviews are welcome.
Mostly, this is just preprocessing gymnastic.
The only thing that worries me a bit is that previous functions like uppercase (and friends) now become uppercase_ascii (and friends).

UnixJunkie:
Unless someone complains, I should merge this next week.
In the meantime, reviews and comments are very welcome.

UnixJunkie:
cc @sim642

UnixJunkie:
Apparently, we only have three unit tests failing:
```
random seed: 1343689
\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\
Failure: >shuffle>src/batArray.mlv:934

src/batArray.mlv:934:  

  let s = Random.State.make [|11|] in 
  let a = [|1;2;3;4;5;6;7;8;9|] in 
  shuffle ~state:s a; 
  a = [|7; 2; 9; 5; 3; 6; 4; 1; 8|]


///////////////////////////////////////////////////////////////////////////////
\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\
Failure: >shuffle>src/batList.mlv:1547

src/batList.mlv:1547:  

  let s = Random.State.make [|11|] in 
  shuffle ~state:s [1;2;3;4;5;6;7;8;9] = [7; 2; 9; 5; 3; 6; 4; 1; 8]


///////////////////////////////////////////////////////////////////////////////
Ran: 2114 tests in: 46.59 seconds.                                        
FAILURE
make: *** [Makefile:236: qtest-byte] Error 1
```

sim642:
> Apparently, we only have two unit tests failing:

AFAIK OCaml 5 changes the standard PRNG algorithm, but the expected results in this test are probably specific to the old one. I guess the expected result could be made conditional, if we want to be checking the exact order at all.

UnixJunkie:
Yes, those two unit tests have been updated now.

UnixJunkie:
Ok, another test is failing:
```
==============================================================================
Error: All:18:BatRandom:2:enum_default_priv

Out of memory
------------------------------------------------------------------------------
Ran: 312 tests in: 19.80 seconds.
FAILED: Cases: 312 Tried: 312 Errors: 1 Failures: 0 Skip:  0 Todo: 0 Timeouts: 0.
make: *** [Makefile:257: testsuite-only-byte] Error 1
```

sim642:
In my PR I didn't investigate it and just disabled, but now looking at it I think it has to do with this:
https://github.com/ocaml-batteries-team/batteries-included/blob/17886db2ae8fc8b9d591ec9f18884b342eb44b25/src/batRandom.mlv#L163-L165
That seems to be assuming and abusing that `Stdlib.Random` internally is `type t = { st : int array; mutable idx : int }` as it's in OCaml 4, but in OCaml 5 it's `type t = (int64, int64_elt, c_layout) Array1.t` instead.
I guess by sheer coincidence that memory representation misuse is not crashing, but subtly putting it into some invalid (infinite-looping?) state.

src/batCharParser.ml
```diff
@@ -66,11 +66,11 @@ let string s = label ("\"" ^ s ^ "\"") (
   )
 
 let case_char c =
-  if BatChar.is_letter c then one_of [Char.uppercase c; Char.lowercase c]
+  if BatChar.is_letter c then one_of [Char.uppercase_ascii c; Char.lowercase_ascii c]
   else char c
 
 let case_string s = label ("case insensitive \"" ^ s ^ "\"") (
-    let s'  = String.lowercase s in
+    let s'  = String.lowercase_ascii s in
```
sim642, line 73: Not sure if it matters to anyone, but `lowercase`→`lowercase_ascii`, etc slightly change behavior on OCaml < 5 as well, because the non-`_ascii` versions used Latin-1.

myocamlbuild.ml
```diff
@@ -9,7 +9,7 @@ module Pack = Ocamlbuild_pack
 
 let ocamlfind x = S[A"ocamlfind"; A x]
 
-let packs = "bigarray,num,str"
+let packs = "camlp-streams,str,unix"
```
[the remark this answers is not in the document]
UnixJunkie: I'll fix this one. Also, I will test backward compatibility.

src/batGc.mliv
```diff
@@ -267,11 +275,9 @@ val allocated_bytes : unit -> float
...
     with [int] on 32-bit machines. *)
 
-##V>=4.3####V<multicore##external get_minor_free : unit -> int = "caml_get_minor_free"
-##V=4.3##          [@@noalloc]
-##V=4.4##          [@@noalloc]
-##V>=4.3####V<multicore##(** Return the current size of the free space inside the minor heap.
-##V>=4.3####V<multicore##    @since 2.5.0 and OCaml 4.03.0 *)
+##V<5####V>=4.3##external get_minor_free : unit -> int = "caml_get_minor_free"  [@@noalloc]
+##V<5####V>=4.3##(** Return the current size of the free space inside the minor heap.
+##V<5####V>=4.3##    @since 2.5.0 and OCaml 4.03.0 *)
```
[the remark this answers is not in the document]
UnixJunkie: Good catch seeing this one. I will correct.

src/batteries.mlv
```diff
@@ -176,20 +176,20 @@ end
...
+##V<5####V>=4.3##  and  type in_channel = Pervasives.in_channel
+##V<5####V>=4.3##  and  type out_channel = Pervasives.out_channel
+##V<5####V>=4.3##  and  type open_flag = Pervasives.open_flag
+##V<5####V>=4.3##  and  type ('a, 'b, 'c, 'd, 'e, 'f) format6 =
+##V<5####V>=4.3##    ('a, 'b, 'c, 'd, 'e, 'f) Pervasives.format6
+##V<5####V>=4.3##  and  type ('a, 'b, 'c, 'd) format4 =
+##V<5####V>=4.3##    ('a, 'b, 'c, 'd) Pervasives.format4
+##V<5####V>=4.3##  and  type ('a, 'b, 'c) format =
+##V<5####V>=4.3##    ('a, 'b, 'c) Pervasives.format
+##V<5####V>=4.3##)
```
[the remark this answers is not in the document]
UnixJunkie: If we just include it, it triggers many errors (already defined modules being redefined again).
But, I am not too worried about this: nobody sane should do 'open Batteries' in their code.
And I think this module is for those people.

src/batCharParser.ml
UnixJunkie: Ok, so  a safer route would be to not change for OCaml < 5 (using preprocessing) but announce for ocaml5 a breaking change for people who were dealing with latin-1 strings.

UnixJunkie:
Thanks to the feedback from @sim642, I could fix BatRandom.
Now, only the tests about interface compatibility are failing.

myocamlbuild.ml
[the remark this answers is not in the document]
UnixJunkie: camlp-streams is now in the opam depends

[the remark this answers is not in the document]
UnixJunkie: the myocamlbuild file should be fixed now

src/batGc.mliv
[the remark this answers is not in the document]
UnixJunkie: I reverted this change

src/batCharParser.ml
UnixJunkie: this is more conservative now (ocaml < 5) is not impacted by the change; using preprocessing

UnixJunkie:
@sim642 the state of the PR takes into account all your comments.
Now, only some interface compatibility tests are broken.
I'll look at those a little later; maybe today.

UnixJunkie:
all tests are passing now. I'll have a fresh look on Monday; unless there are more comments, I will merge then.
@sim642 if you want to have a look, I think the PR should be pretty much final now.

batteries.opam
```diff
@@ -19,7 +19,8 @@ build: [
 install: [make "install"]
 run-test: [make "test"]
 depends: [
-  "ocaml" {>= "4.00.0" & < "4.15.0"}
+  "ocaml" {>= "5.00.0"}
```
sim642, line 22: Is this lower bound supposed to be temporary? Because if it isn't, then there's no point to any of the conditional preprocessing.

src/batRandom.mliv
```diff
@@ -206,6 +206,15 @@ module State : sig
...
 ##V>=4.14##val bits64 : t -> Int64.t
 ##V>=4.14##val nativebits : t -> Nativeint.t
+
+##V>=5##val split : t -> t
+##V>=5##(** Draw a fresh PRNG state from the given PRNG state.
+##V>=5##    (The given PRNG state is modified.)
+##V>=5##    The new PRNG is statistically independent from the given PRNG.
+##V>=5##    Data can be drawn from both PRNGs, in any order, without risk of
+##V>=5##    correlation.  Both PRNGs can be split later, arbitrarily many times.
+##V>=5##    @since 5.0.0 *)
```
sim642, line 216: Minor nitpick: I think other `@since` tags are explicit about whether they talk about Batteries or OCaml version. Same for other OCaml 5 added functions.

src/batteries.mlv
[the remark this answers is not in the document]
sim642: > nobody sane should do 'open Batteries' in their code.
> And I think this module is for those people.

Heh, I have a project that occasionally opens Batteries (which ideally could be refactored away). I should try it out with this now to see if there's any downside.

batteries.opam
UnixJunkie: What I wanted to say in the opam file is: if ocaml >= 5 then depend on camlp-streams.
But, I don't know how to say that exactly.

src/batRandom.mliv
UnixJunkie: I also thought about it. I might update those tags on Monday.

batteries.opam
UnixJunkie: @kit-ty-kate Dear Kate, do you know how to say this to opam: "if ocaml >= 5 then depend on camlp-streams" ?

UnixJunkie:
@sim642 You can opam pin this branch of batteries and test. I guess your 'open Batteries' is still working.
I think, opening a whole module globally is the worst thing to do in some OCaml code (and might render your code as unreadable
as Haskell code).
The _only_ module for which I allow an exception is Printf, because all Printf's functions are called printf or the like (sprintf, etc).
Module short aliases are way better than global open:
```ocaml
module L = List
```
Also, there are also several local open options (let open M in or M.(...)).
so that other programmers understand very explicitly where all functions come from.

batteries.opam
sim642: Maybe something like this:
```
  "ocaml" {>= "4.00.0"}
  ("ocaml" {< "5.0"} | "camlp-streams")
```
There are variations of this in some opam-repository packages at least.

kit-ty-kate: yeah something like that would work though i would encourage you to always use camlp-streams for simplicity sake. On OCaml < 5.0 the module is already a simple alias to `Stdlib.Stream`

batteries.opam
```diff
@@ -19,7 +19,8 @@ build: [
 install: [make "install"]
 run-test: [make "test"]
 depends: [
-  "ocaml" {>= "4.00.0" & < "4.15.0"}
+  "ocaml" {>= "4.00.0"}
+  ("ocaml" {< "5.0"} | "camlp-streams")
```
kit-ty-kate, line 23:
```suggestion
  "camlp-streams"
```

myocamlbuild.ml
```diff
@@ -9,15 +9,25 @@ module Pack = Ocamlbuild_pack
...
 
-let packs = "bigarray,num,str"
+let ocaml_major_version =
+  int_of_char Sys.ocaml_version.[0]
+
+let packs =
+  if ocaml_major_version <= 4 then
+    "bigarray,num,str"
+  else
+    "num,str,camlp-streams,unix"
```
kit-ty-kate, line 19:
```suggestion
let packs = "num,str,camlp-streams,unix"
```

kit-ty-kate (commented):
It would be much simpler this way to avoid having to generate the META file with some more fill up in configure. It's not worth the effort given the community is already using camlp-streams unconditionally and `Stream` = `Stdlib.Stream` in OCaml < 5.0 already in camlp-streams

batteries.opam
sim642: camlp-streams has ocaml lower bound 4.02.3, but batteries currently has 4.00.0. If we accept that minimum bump, then it'd be cleaner indeed.

sim642:
> You can opam pin this branch of batteries and test

Tried it out, but apparently opam runs `setup.ml`, which doesn't compile:
```
#=== ERROR while compiling batteries.3.5.1 ====================================#
# context     2.0.10 | linux/x86_64 | ocaml-variants.5.0.0+trunk | pinned(git+file:///home/opam/batteries-included#ocaml5_WIP_01#8bddf643)
# path        ~/.opam/5.0/.opam-switch/build/batteries.3.5.1
# command     ~/.opam/5.0/bin/ocaml setup.ml -configure --prefix /home/opam/.opam/5.0
# exit-code   2
# env-file    ~/.opam/log/batteries-101-c33e1c.env
# output-file ~/.opam/log/batteries-101-c33e1c.out
### output ###
# File "./setup.ml", line 575, characters 4-15:
# 575 |     Stream.from next
#           ^^^^^^^^^^^
# Error: Unbound module Stream
```

I had to manually installing from this PR branch and bypass that to just build/install with `make`. Other than that, this works perfectly with my project (other than a single `String.lowercase` in there)!

UnixJunkie:
I'll try to look at all the comments today, take them into account into some way then merge the PR.
I am OK to bump the lbound for batteries, since this will be a major release for ocaml-5.

batteries.opam
UnixJunkie: we should bump the ocaml lower-bound then

UnixJunkie: in addition to depending on camlp-streams

UnixJunkie:
@sim642 we should get rid of this setup.ml file. This is because batteries still depend on oasis currently (setup.ml is generated by oasis). I don't know why someone introduced this one more layer of complexity of the build process.
I don't know how to get rid of it for good.
If you have ideas, they are very welcome; cf. https://github.com/ocaml-batteries-team/batteries-included/issues/906

batteries.opam
UnixJunkie: I went with @kit-ty-kate 's suggestion

src/batRandom.mliv
UnixJunkie: this has been dealt with

PR #1087, opened 2022-08-29
change the way OCaml5's Stdlib module is supported (replacing Pervasives)

UnixJunkie:
I try to not put all files under pre-processing to minimize the change.
The previous way it was managed was not compiling for ocaml<5.

### DIFF
--- a/src/batArray.mlv
+++ b/src/batArray.mlv
@@ -22,7 +22,11 @@
 
 include Array
 
-##V>=5## module Pervasives = Stdlib
+##V>=5##module Pervasives = Stdlib
+
+##V>=5##(*$inject
+##V>=5##  module Pervasives = Stdlib
+##V>=5##*)
 
 type 'a t = 'a array
 type 'a enumerable = 'a t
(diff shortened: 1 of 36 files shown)


UnixJunkie:
@sim642 I had to change the way the new Stdlib module from ocaml5 is handled.
Because the previous way was not compiling with ocaml < 5.

UnixJunkie:
I feel only the changes in BatHeap and BatInt might need to be reviewed (because they have a risk of introducing bugs).
All the rest is just trivial.

UnixJunkie:
Now, all tests are passing on both ocaml 4.02.3 and 5.0.0~alpha1.
This is a lot of small stupid changes.

sim642:
> I feel only the changes in BatHeap and BatInt might need to be reviewed (because they have a risk of introducing bugs).
> All the rest is just trivial.

According to git blame, a few of those touch the code from #278, but that fix remains, so it looks fine.

Was about to ask about all the `$inject` stuff, but apparently that's a qtest thing.

UnixJunkie:
Yes, inject is to make qtest aware of some global things.

UnixJunkie:
Thanks for the review.

PR #1089, opened 2022-09-05
String split on string comp bug 01

UnixJunkie:


UnixJunkie:
@W95Psp you are welcome to review it

PR #1091, opened 2022-10-05
Fix BatBitset.inter

### LINKED ISSUE #1090: result of BatBitSet.inter(and intersect) is not correct in some cases
`let s1 = BatBitSet.create_full 16 (* represents {0, 1, ..., 14, 15} *)`
`let s2 = BatBitSet.create_full 8 (* represents {0, 1, ..., 6, 7} *)`
`let s3 = BatBitSet.inter s1 s2 (* expected: {0, 1, ..., 6, 7} *)`
`let _ = print_int (BatBitSet.count s3) (* expected: 8, actual: 16 *)`

The helper function `apply_set_op` may need some extra post-processing for `Inter` operation(for example, `if op = Inter && len1 > len2 then ...`)

gasche:
This hopefully fixes #1090. cc @yhyoon and @UnixJunkie

### DIFF
--- a/src/batBitSet.ml
+++ b/src/batBitSet.ml
@@ -323,6 +323,7 @@ let apply_set_op op t1 t2 =
   let len1 = Bytes.length !t1 in
   let len2 = Bytes.length !t2 in
   let clen = min len1 len2 in
+  (* this operates on the common substring only *)
   while !idx < clen do
     let c1 = Char.code (Bytes.unsafe_get !t1 !idx) in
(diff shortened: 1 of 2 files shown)


UnixJunkie:
Match op inside the loop: is this smart?
Shoudn't the proper function be defined before the loop?

UnixJunkie:
I am sorry but it is quite late for me in here to review code.
I don't even understand why the previous code had only tests for len1 < len2.

PR #1092, opened 2022-10-13
Master

UnixJunkie:


UnixJunkie:
BatBitSet.apply_set_op: don't match inside the loop
plus changelog update

PR #1095, opened 2022-10-25
Remove unused oasis

### LINKED ISSUE #906: rm the _oasis file

sim642:
Closes #906. Hopefully final piece for #1081.

The `ocaml setup.ml -configure` that the opam file was executing only generated `setup.data`, which is used for nothing, and nothing else.

The `_oasis` file is thus also essentially unused: only a couple of build scripts grep the batteries version number from there, so I kept just that portion.
~~Should we maybe rename this fake `_oasis` to `_version` or something?~~ Simplified to `VERSION`.

The META file generation, mentioned in #906, doesn't actually use oasis in any meaningful way. `mkconf` just substitutes the version number grepped from `_oasis`.


kit-ty-kate:
I've open a PR against this branch to remove the _oasis file entirely: https://github.com/sim642/batteries-included/pull/1

kit-ty-kate:
once this is merged i think batteries can be released as the tests work just fine with 5.0.0~beta1 and the final release is closing in fast.

kit-ty-kate:
oops

UnixJunkie:
Thanks a lot for having looked into this.
The only worry I had was if the ocamldoc can still be built.

UnixJunkie:
```bash
# make doc
ocamlbuild -no-links -use-ocamlfind batteries.docdir/index.html
Finished, 0 targets (0 cached) in 00:00:00.
Solver failed:
  Ocamlbuild knows of no rules that apply to a target named benchsuite/lib/Pervasives.ml. This can happen if you ask Ocamlbuild to build a target with the wrong extension (e.g. .opt instead of .native) or if the source files live in directories that have not been specified as include directories.
Compilation unsuccessful after building 321 targets (312 cached) in 00:00:00.
make: *** [Makefile:120: doc] Error 6
```

UnixJunkie:
This is a little annoying.
But maybe not even broken by this PR.

UnixJunkie:
At tag v3.5.1; 'make doc' was still working.
I don't have time to investigate right now.

PR #1098, opened 2022-12-27
Fix `dune build`

sim642:
First of a series of patches towards #738.

Issue #738, opened 2017-03-16
switch to dune instead of ocamlbuild


UnixJunkie:
if it parallelizes the build and works on windows, that would be quite nice.

gasche:
I believe that ocamlbuild works on Windows but yes, feel free to try jbuilder -- we do need to make sure that it does also work on Windows, I think.

UnixJunkie:
we should really give it a try. I guess it would parallelize the build quite well.

c-cube:
Just my 2 cents: on containers, dune makes the build fast and parallelizes well. It also allows qtest to be used with a few lines in jbuild files.

gasche:
I'm sure it would be nice to have a dune-based build system for Batteries, and it should not be too complicated -- we're not doing complex build stuff. The question is rather of who is motivated to implement it. (It might be simpler to move to `cppo` than to preserve the `prefilter` stuff, which is probably a good idea.)

(What's the bound on OCaml versions that comes with adopting dune? Should we keep an ocamlbuild or Makefile on the side for older OCaml versions?)

c-cube:
the bound is >= 4.02.3 in practice, I think. It's kind of annoying for me too, but it's a tradeoff to make… :/

UnixJunkie:
I'd also like that we move to cppo. I'd also like that all files are .ml/.mli. I don't see why some files have to be called .mlv/.mliv. It's just annoying when you edit the wrong file, and its not uniform.
The build system should know when to preprocess a file or not.
And, if all files are preprocessed, a preprocessor is something that is supposed to be lightning fast.

UnixJunkie:
If we switch to jbuilder, this may send a signal to some people that at least part of the open-source ocaml community is seriously considering jbuilder, and it may incent more people to switch to it.
I think it is very positive for the ocaml ecosystem if the community starts to move "en masse" to such a build system. The previous bazaar of build systems in ocaml was a problem for newcomers, and the bad taste of the build system that was shipping with the compiler for years was also problematic (though this is maybe just my personal opinion).

UnixJunkie:
Some venerable OCaml programmers have not switched yet...

UnixJunkie:
I haven't switched all my projects yet also, but this might happen over time.

XVilka:
[...]

PR #1099, opened 2022-12-27
Fix `dune runtest`

sim642:
Follow-up to #1098 and towards #738.

The main thing is that dune's `inline_tests` runs the qtest backend on the unpreprocessed files (probably because other inline test backends are based on ppx). The solution is to first run `qtest extract` and pipe the result through `prefilter`.
This order is different from Makefile/ocamlbuild, thus the `$inject` changes to move `##V>=5##` only inside. Luckily only parts that are affected are from #1087.


UnixJunkie:
If I run 'make test', it tells me 2115 tests were ran and OK, then 313 tests ran and OK.

UnixJunkie:
'dune runtest' on the other hand, just tells me 313 tests were ran and OK.

UnixJunkie:
Currently, 'dune runtest' also produces an error:
```
File "src/.batteries.inline-tests/inline_test_runner_batteries.ml-gen", line 13, characters 20-30:
13 | module Pervasives = Pervasives[@warning "-3"]
                         ^^^^^^^^^^
Error (warning 49 [no-cmi-file]): no cmi file was found in path for module Pervasives
```
Can this be corrected?

sim642:
> If I run 'make test', it tells me 2115 tests were ran and OK, then 313 tests ran and OK.

> 'dune runtest' on the other hand, just tells me 313 tests were ran and OK.

That might be due to dune not rerunning things without changes. On a clean run I do get both:

```console
simmo@roentgenium:~/dev/batteries-included (dune-runtest)$ dune clean
simmo@roentgenium:~/dev/batteries-included (dune-runtest)$ dune runtest
.........................................................................................................................................................................................................................................................................................................................
Ran: 313 tests in: 4.02 seconds.
OK
random seed: 213784516                 
Ran: 2115 tests in: 8.77 seconds.                                              
SUCCESS
```

sim642:
> Currently, 'dune runtest' also produces an error:
> 
> ```
> File "src/.batteries.inline-tests/inline_test_runner_batteries.ml-gen", line 13, characters 20-30:
> 13 | module Pervasives = Pervasives[@warning "-3"]
>                          ^^^^^^^^^^
> Error (warning 49 [no-cmi-file]): no cmi file was found in path for module Pervasives
> ```
> 
> Can this be corrected?

Good catch! I forgot to check everything on OCaml 5, but it should be fixed now.

UnixJunkie:
thanks for the update and additional fixes.
'dune runtest' is about four times faster than 'make test' !
Both from a fresh checkout, with nothing compiled.

UnixJunkie:
Also, both targets are still working, which is reassuring.

UnixJunkie:
'make test' AND 'dune runtest' are supposed to work for ocaml<5 and ocaml>=5.
If you find that it is not the case, please tell, or even better, send a PR with a fix. :)

I guess that in the future, we might drop the Makefile.
But, for that, we would need dune to install and do everything fine for the lowest ocaml version we support.

PR #1104, opened 2023-02-08
Fixes to allow batteries to be vendored in other packages

gridbugs:
This fixes two problems I ran into when trying to use opam-monorepo to vendor batteries inside a monorepo project.


UnixJunkie:
Thanks for the proposal.
I will test and merge if OK.

UnixJunkie:
Wasn't there already an unthreaded version of batteries that could be installed before (via make)?
Are we keeping the same name for it?

UnixJunkie:
Ok, nothing seems broken so merging.
The previous name was also batteries.unthreaded (in META.in).
Thanks a lot!

PR #1105, opened 2023-02-09
Expose batteries.top library via dune and other dune library changes

sim642:
#1104 did the unthreaded vs threaded library split in dune, which I had also planned on. But here's a few extra changes to polish the setup.

### Changes
1. Expose batteries.top library. This is currently different from the ocamlbuild setup, but it's not possible to have dune produce special `toploop` stuff into `META` automatically. Also, many other libraries have a top sublibrary, so it's a sensible pattern to go for.
2. Use prefilter and qtest also for threaded batteries in dune (#1104).
4. Use current `META.in` descriptions for dune libraries. Just to get the dune build as close to the ocamlbuild one.
5. Fix prefilter path in `batteries_compattest` a la #1104.

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -1,8 +1,8 @@
 (library
  (name batteries_unthreaded)
  (public_name batteries.unthreaded)
- (synopsis "A community-maintained standard library extension")
- (modules (:standard \ ("batteries_compattest" batteriesThread batRMutex batMutex)))
+ (synopsis "Batteries Included (for use in unthreaded programs)")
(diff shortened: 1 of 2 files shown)


UnixJunkie:
I will test then merge.

What a fight over this dune file. :)
It would be nice if this file stays maintainable; i.e. doesn't grow like crazy and does not start to use all the bells and whistles from dune.

UnixJunkie:
Thanks a lot Simmo for your many contributions.

PR #1106, opened 2023-02-15
Add GitHub Actions CI

sim642:
Currently there's no CI but this is valuable to have. This checks all supported OCaml versions, right not both ocamlbuild and dune.

This includes some small fixes such that the dune build passes on all OCaml versions.

### TODO
- [x] Remove unintended whitespace changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,12 +31,15 @@ jobs:
           - true
 
         include:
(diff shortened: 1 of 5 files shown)


UnixJunkie:
Thanks for this contribution.
I hope this will not be too brittle and require a lot of maintenance.

Issue #1101, opened 2023-01-17
get rid of ocamlbuild


UnixJunkie:

UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/658
and the closed issue https://github.com/ocaml-batteries-team/batteries-included/issues/738

UnixJunkie:
related to https://github.com/ocaml-batteries-team/batteries-included/issues/1102#issuecomment-1408302630

PR #1109, opened 2023-03-20
Generate opam file from dune

sim642:
Towards #738 and #1101.

Adds the package definition into `dune-project` based on data from the currently hand-written `batteries.opam`. Then uses dune to generate the opam file.

The dune-generated opam file has one major difference: **the dune build becomes the default** and ocamlbuild is no longer required.
All the ocamlbuild stuff is still around. If we commit to this change, I can make a follow-up PR to remove all of that for good.


UnixJunkie:
Thanks for this.
You should add your name in the list of maintainers.

PR #1110, opened 2023-03-22
Add Simmo Saan to maintainers, remove ocamlbuild CI job

sim642:
Small follow-up to #1109.

The CI job with `make` is failing because the ocamlbuild dependency is no longer declared and installed. Once I get fully rid of ocamlbuild stuff, we could have a much smaller Makefile that just does respective dune calls.


UnixJunkie:
thanks!

PR #1111, opened 2023-05-25
BatString: fix possible stack overflow on split_on_string_comp

mtzguido:
The exception handler should not cover the recursive call.

--

Hello, I noticed that `split_on_string` would stack overflow for long strings, like so:
```
$ ocaml
        OCaml version 4.13.0

      _______________________________
    [| +   | |   Batteries 3.6.0  - |
     |_____|_|______________________|
      _______________________________
     | -  Type '#help;;'      | | + |]
     |________________________|_|___|


# BatString.split_on_string " " (BatString.repeat " " 10000000);;
Warning 6 [labels-omitted]: label by was omitted in the application of this function.
Stack overflow during evaluation (looping recursion?).
```
Seems to me like this is a sensible fix for it. Maybe a test can also be added to the list below?

Thanks!

### DIFF
--- a/src/batString.mlv
+++ b/src/batString.mlv
@@ -343,6 +343,16 @@ let tail s pos =
   let slen = length s in
   if pos >= slen then "" else sub s pos (slen - pos)
 
+let repeat s n =
+  let buf = Buffer.create ( n * (String.length s) ) in
+  for _i = 1 to n do Buffer.add_string buf s done;
+  Buffer.contents buf
+(*$T repeat
+   repeat "fo" 4 = "fofofofo"
+   repeat "fo" 0 = ""
+   repeat "" 4 = ""
+*)
+
 (*$T left
   left "abc" 1 = "a"
   left "ab" 3 = "ab"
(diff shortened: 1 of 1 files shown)


UnixJunkie:
thanks a lot, I will test your proposed fix and then merge

UnixJunkie:
@mtzguido in your PR, could you add a unit test to check that indeed your test case is not raising an exception anymore?

UnixJunkie:
Your PR is fine; just a new unit test is welcome.
I will probably reformat your code once you are done but don't bother about that.

mtzguido:
Thanks for taking a look, François!

I just added a test that catches the problem. Though I did move `repeat` higher up in the file to write it. I suppose that's fine...? In any case feel free to rejig things as you see fit!

Thanks,
Guido

mtzguido:
(I have to say I don't particularly know if ocaml programs do anything special to their stack limit.. but that test does reproduce the problem in my setup. To pass that test I think that at the very least one needs an 80MB stack, so unless the `ulimit` was changed, I don't think it would pass anywhere.)

UnixJunkie:
Probably janestreet does change the stack size limit on their servers, but I guess most people don't.

UnixJunkie:
I'll probably merge your PR this afternoon.

UnixJunkie:
We will probably not include the unit test in the end; too long to trigger.

UnixJunkie:
Which ocaml version are you using?
How long to trigger the bug?

UnixJunkie:
superseded by https://github.com/ocaml-batteries-team/batteries-included/pull/1112/files

UnixJunkie:
I tagged v3.6.1 master w/ your proposed fix included.

mtzguido:
Thank you!

UnixJunkie:
@mtzguido did you trigger this bug in the wild?
Or was it found by a static analysis tool?
If it was found by an automated tool, then performing the analysis on the whole batteries code is very welcome.

mtzguido:
I just triggered it, I got a stack overflow while trying to split a rather big file into lines and tracked it down to this.

mtzguido:
Hi @UnixJunkie, seems like the new release did not make it into opam? Should a PR be made into https://github.com/ocaml/opam-repository?

UnixJunkie:
This is not a proper release, I just tagged the bugfix so that people can
opam pin to it.

PR #1112, opened 2023-05-26
split_on_string_comp stack overflow

UnixJunkie:


src/batString.mlv
```diff
@@ -406,12 +406,15 @@ let split_on_string_comp ?(on_empty=[""]) ~by:sep str =
       if curr = len then
         List.rev ("" :: acc) (* for compatibility with old behavior *)
       else
-        try
-          let i = find_from str curr sep in
+        let maybe_i =
+          try Some (find_from str curr sep)
+          with Not_found -> None in
+        match maybe_i with
+        | Some i ->
```
gasche, line 413: Passerby note: the idiomatic way to write this code nowadays is

```ocaml
match find_from str curr sep with
| exception Not_found -> ...
| i -> ...
```

UnixJunkie:
But would this compile on all ocaml versions we support?

Issue #1093, opened 2022-10-20
Switch to odoc?


UnixJunkie:
Instead of ocamldoc.

Should we?

Can this be done automatically? I.e. update all the ocamldoc comments, if needed.

UnixJunkie:
It would be nice if someone with odoc experience can comment.

sim642:
With `dune build @doc`, it works out of the box. There's just a bunch of warnings about heading levels and custom tags.
It would make sense to make this switch together with #738.

PR #1113, opened 2023-06-03
Remove ocamlbuild

### LINKED ISSUE #658: get rid of the .mlv/.mliv files

### LINKED ISSUE #1101: get rid of ocamlbuild

sim642:
In #1109 dune became the default build tool, so this removes most of the ocamlbuild stuff. Closes #1101.

With ocamlbuild compatibility no longer needed, preprocessing is also simplified. Closes #658.

There's still a bunch of old ocamlbuild-related things in the `build` directory but I think they weren't even used before. So I didn't do a full clean-up because I'm not sure what all of it is and has been used for.

I did these things a long time ago, but never opened a PR. Not sure if I still intended to do something.


UnixJunkie:
I will review and test it.
If that works, next step I would like to rename all .mlv/mliv files to .ml/mli.
This specific extension has _always_ annoyed me.

UnixJunkie:
You already did that! Yummy yummy.

UnixJunkie:
The test_install target in the Makefile was useful.

So, dune still cannot compile an exe with a one-liner invocation?
The only feature ocamlbuild can be praised for...

UnixJunkie:
Our test suite does not run w/ ocaml 4.02.3, but I think this is not new.

UnixJunkie:
'make doc' does not work in this branch.
It produces many errors.

UnixJunkie:
Apparently, make doc should be preceded by make.
Strange this is not already encoded in the Makefile.

UnixJunkie:
@sim642 can you check the 'make doc' target is still doing what it should.
I merged; but I am not sure about this one.

sim642:
> The test_install target in the Makefile was useful.
> 
> So, dune still cannot compile an exe with a one-liner invocation? The only feature ocamlbuild can be praised for...

It should be able to. I might've just not noticed the test_install target being different from all the other testing targets. I'll try to add it back.


> make doc' does not work in this branch (I tried under ocaml-5.0.0).
> It produces many errors.



> Apparently, make doc should be preceded by make.
> Strange this is not already encoded in the Makefile.

I recall many odoc warnings not really being an issue but I'll look into it for #1093. I have some old branch also related to fixing some odoc warnings, so I should dig that up. And I think dune should be able to build the docs without explicitly building the code before.

UnixJunkie:
It is possible the odoc errors are just warnings; but I am not sure and it is still annoying.
If our doc is meant to be generated by odoc from now on, then we should use it properly.
Throwing many warnings might hide some things which are actually errors; people should not get used to living with
code throwing many warnings.

sim642:
> The test_install target in the Makefile was useful.
> 
> So, dune still cannot compile an exe with a one-liner invocation? The only feature ocamlbuild can be praised for...

I looked into it now and see what you meant. Indeed I think it also requires a very basic `dune` file.

But actually with that, dune with its support for vendoring and monorepos will actually find the batteries library from the workspace even when it's not installed into the opam switch. So it wouldn't test what the script is supposed to be testing. I guess it can just remain like that for now.

UnixJunkie:
Yes, I uncommented this test_install make target (that you previously commented out).
So, it behaves as before and still uses ocamlbuild for the moment.
But, since this is only useful to people who run some tests before a batteries release, no big deal.

PR #1121, opened 2023-09-09
Add workaround for missing batteriesConfig dependency in inline tests

sim642:
Hopefully fixes this issue: https://github.com/ocaml/opam-repository/pull/24327#issuecomment-1709839625.

I could reproduce it by commenting out the `test` stanza in `src/dune` and running `dune clean && dune build @src/runtest`. After these changes that problem no longer appears for me like that, so hopefully it's also fixed for the CI.

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -9,7 +9,8 @@
  (libraries num str camlp-streams unix bigarray)
  (inline_tests
    (backend qtest_batteries)
-   (deps %{project_root}/qtest/qtest_preamble.ml)
+   (deps %{project_root}/qtest/qtest_preamble.ml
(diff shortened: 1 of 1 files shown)


UnixJunkie:
```bash
dune clean && dune build @src/runtest
```
does reproduce the error, even w/ ocaml-5.0.0

UnixJunkie:
You only fixed the target batteries_unthreaded; in src/dune there is also batteries which also have this same inline_tests section.
Is this voluntary?

sim642:
> You only fixed the target batteries_unthreaded; in src/dune there is also batteries which also have this same inline_tests section.
> Is this voluntary?

But that only has `(modules batteriesThread batRMutex batMutex)`, so `BatteriesConfig` shouldn't belong to its inline tests.

UnixJunkie:
Ok, I see.
Thanks for the fix.

PR #1128, opened 2024-02-27
Add OCaml 5.2 support

### LINKED ISSUE #1127: OCaml 5.2 support
The current version of batteries is not compatible with OCaml 5.2 (currently released as 5.2.0\~alpha1)

kit-ty-kate:
Fixes #1127


UnixJunkie:
thanks a lot; reviewing is way faster than having to do it myself

PR #1124, opened 2023-11-26
Add support for OCaml 5.1.1

kit-ty-kate:
Fixes #1123

PR #1131, opened 2024-04-14
Fix version equality without micro version in prefilter

sim642:
Fixes regression from #1124. Since that PR, GitHub Actions CI has been failing on OCaml ≤ 4.04.

/cc @kit-ty-kate


UnixJunkie:
thanks a lot

PR #1132, opened 2024-05-04
Add Stdlib type equality to Set.Make, Map.Make and Hashtbl.Make

sim642:
For most types batteries already has such type equalities, but crucially for `BatSet.Make`, `BatMap.Make` and `BatHashtbl.Make` they were missing. Containers has the same type equalities exposed.

These are particularly useful for when batteries lags behind `Stdlib` (which can often happen nowadays) and one wants to use some functions from batteries, but also new ones from `Stdlib`. This type equality makes it at least possible without having to immediately patch batteries every time and wait for a release.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+## NEXT_RELEASE
+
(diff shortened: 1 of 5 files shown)


UnixJunkie:
thanks a lot!

PR #1133, opened 2024-05-04
Add reset, find_opt and Seq functions to BatHashtbl

sim642:
These are in Stdlib's `Hashtbl`, but have been missing from batteries. The `Seq` functions are just delegated from Stdlib only on OCaml >= 4.07 where they were added. I don't think it's worth reimplementing them for all OCaml versions at this point.


UnixJunkie:
Thanks!

PR #1137, opened 2024-12-31
Update GitHub Actions CI

sim642:
The CI on ee91f7899976584c87d92c3c1363e0a748cd9a9c fails miserably because ubuntu-latest is now 24.04 and doesn't have darcs. The new version of setup-ocaml fixes the problem.

Updating OCaml versions in the CI while at it.


UnixJunkie:
thanks for taking care of the CI

PR #1139, opened 2025-07-02
Support OCaml 5.4

tmcgilchrist:
Initial hacking to support OCaml 5.4. 

I mechanically added pieces until it compiled, so there's some parts that are questionable. In particular:
 1. Generated `batteries.opam` differs from the checked in version 
 2. Installing batteries with OCaml 5.4 + options-fp fails
 3. Documentation strings differ slightly from the upstream ocaml
 4. GitHub CI setup includes older OCaml versions that are no longer supported (< 4.05)
 
CI results https://github.com/tmcgilchrist/batteries-included/pull/1/checks


sim642:
> 1. Generated `batteries.opam` differs from the checked in version

I noted in https://github.com/ocaml-batteries-team/batteries-included/commit/3895a849cc398b9d6472e027f90101024e982218#commitcomment-150951031 that the opam-repository one might've been out of date and the generated one should be right. But I never got around to checking it properly.

src/batFormat.ml
```diff
@@ -56,12 +67,28 @@ let formatter_of_output out =
...
+(* ##V<5## ~spaces:(spaces_of out); *)
+(* ##V>=5## pp_set_formatter_out_functions f *)
+(* ##V>=5## { out_string = output; *)
+(* ##V>=5.4##   out_width = utf8_scalar_width; *)
+(* ##V>=5##   out_flush = flush; *)
+(* ##V>=5##   out_newline = newline_of out; *)
+(* ##V>=5##   out_spaces = spaces_of out; *)
+(* ##V>=5##   out_indent = spaces_of out }; *)
+(*   f *)
+
```
sim642, line 91: Is this a leftover from something?

src/batResult.ml
```diff
@@ -21,12 +21,15 @@ let ok v = Ok v
 let error e = Error e
 let value r ~default = match r with Ok v -> v | Error _ -> default
 let get_ok = function Ok v -> v | Error _ -> invalid_arg "result is Error _"
+##V>=5.4##let get_ok' = function Ok v -> v | Error e -> invalid_arg e
```
sim642, line 24: Given that these are implemented here anyway, is there really a reason to add the version constraint? A la backport.

src/batString.mli
```diff
@@ -1072,11 +1120,11 @@ module NumString : BatInterfaces.OrderedType with type t = t
...
-val edit_distance : t -> t -> int
-(** Edition distance (also known as "Levenshtein distance").
-    See {{:http://en.wikipedia.org/wiki/Levenshtein_distance} wikipedia}
-    @since 2.2.0
-*)
+##V<5.4##val edit_distance : t -> t -> int
+##V<5.4##(** Edition distance (also known as "Levenshtein distance").
+##V<5.4##    See {{:http://en.wikipedia.org/wiki/Levenshtein_distance} wikipedia}
+##V<5.4##    @since 2.2.0
+##V<5.4##*)
```
sim642, line 1127: I'm not sure how good of an idea it is to have the behavior of this function change based on OCaml version, especially because the documentation in no way reflects this.
I think somewhere else batteries has things like `exists_stdlib` to deal with such incompatibilities.

src/batSys.mli
```diff
@@ -181,124 +192,198 @@ val max_array_length : int
...
 
+##V>=5.4## type signal = int
+##V>=5.4## (** The type for signal numbers.
+##V>=5.4## 
+##V>=5.4##   Negative numbers are used by OCaml to provide a platform-independent
+##V>=5.4##   number for signals recognised by OCaml. Positive numbers are always the
+##V>=5.4##   platform-dependent value for a given signal.
+##V>=5.4##   The function {!signal_of_int} converts known platform-dependent numbers
+##V>=5.4##   to independent ones, and {!signal_to_int} does the reverse.
+##V>=5.4##   @since 5.4 *)
```
sim642, line 203: Could something break if this type alias was introduced unconditionally for older OCaml versions as well?

sim642:
> 4\. GitHub CI setup includes older OCaml versions that are no longer supported (< 4.05)

I think it makes sense to just remove them since compatible versions of "num" have been archived on opam-repository.

UnixJunkie:
Thanks for contributing. Sorry but I merged the PR from another contributor which I already know and trust.

tmcgilchrist:
No worries, happy to have OCaml 5.4 :-)

PR #1140, opened 2025-08-09
Add support for OCaml 5.4

kit-ty-kate:
I've also remove the opam file generation because the two opam files (real and generated) do not match and i don't have the patience to workaround all the generation's quirks nor do i think having it serves anything. If you disagree feel free to remove that commit

### DIFF
--- a/src/batteries_compattest.ml
+++ b/src/batteries_compattest.ml
@@ -54,6 +54,7 @@ module Stdlib_verifications = struct
 ##V>=4.14## end = struct
 ##V>=4.14##   include BatSeq
 ##V>=4.14##   let equal = equal_stdlib
+##V>=5.4##    let filteri = Stdlib.Seq.filteri
 ##V>=4.14## end
   module Marshal =
     (Marshal: sig
@@ -86,6 +87,7 @@ module Stdlib_verifications = struct
     let starts_with = starts_with_stdlib
     let ends_with = ends_with_stdlib
     let exists = exists_stdlib
+##V>=5.4##let edit_distance = Stdlib.String.edit_distance
   end
   module Sys = (Sys : module type of Legacy.Sys)
   module Unix =
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,8 @@ jobs:
         os:
           - ubuntu-latest
         ocaml-compiler:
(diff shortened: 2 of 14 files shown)


UnixJunkie:
I am very glad that we don't rely on dune for the opam-file generation:
- dune should do one thing and do it well; not try to replace all ecosystem tools
- I don't see the advantage of having to write the whole opam-file using dune's very different syntax that I don't even know
- it will make it easier to check our opam file is synchronized w/ the one in opam-repository

UnixJunkie:
I will try to have a look at this PR over the week-end

UnixJunkie:
If another maintainer wants to have a look at it and merge, please go on, I am no more in academia...

UnixJunkie:
Just a question: is this PR somehow related to this one: https://github.com/ocaml-batteries-team/batteries-included/pull/1139
I prefer to review the one from @kit-ty-kate , because you are a long term contributor so I trust you more.

kit-ty-kate:
Yes both do the same thing roughly. I saw #1139 only after i had opened the PR. The main differences between the two are:
- He copied the documentation from OCaml vs. I didn't as i think it's superflous and too subject to future change to be kept in sync
- He made `BatString.edit_distrance` change behaviour between 5.3 and 5.4 vs i kept the same function between the two and simply ignored the one from stdlib

src/batteries_compattest.ml
```diff
@@ -54,6 +54,7 @@ module Stdlib_verifications = struct
 ##V>=4.14## end = struct
 ##V>=4.14##   include BatSeq
 ##V>=4.14##   let equal = equal_stdlib
+##V>=5.4##    let filteri = Stdlib.Seq.filteri
```
sim642, line 57: Is there a reason to have it here just for compatibility, but not in `BatSeq` itself?

kit-ty-kate: I'm not a user of batteries, i'm only here to upgrade the compatibility of the OCaml ecosystem to the new versions of the compiler (work paid by the OCaml Software Foundation), so i only do the minimum for it to be compatible. If you think this is better, feel free to open a PR queued on this one that does what you want

kit-ty-kate:
Thanks a lot! Would it be possible to get a release? I'm able to handle the opam-repository part if needed.

PR #1143, opened 2025-11-15
Add Array.slide to generate a sliding windows of array

Sumandora:
A simple algorithm that comes in handy every now and then. Would love to see it be a part of Batteries.

obligatory link, although I don't think its needed: https://www.freecodecamp.org/news/sliding-window-technique/

This is my first PR to this project, please tell me if I f*cked something up

### DIFF
--- a/src/batArray.ml
+++ b/src/batArray.ml
@@ -1023,6 +1023,22 @@ let tail a pos = let alen = length a in
       *)
 
 
+let slide n a =
+  if n < 1 then invalid_arg "Array.slide";
+  let length = length a - n + 1 in
+  if length < 1 then invalid_arg "Array.slide";
+  init length (fun i -> init n (fun j -> a.(i + j)))
+
+(*$T slide
+  slide 2 [| 1; 2 |]           = [| [| 1; 2 |] |]
+  slide 2 [| 1; 2; 3 |]        = [| [| 1; 2 |]; [| 2; 3 |] |]
+  slide 2 [| 1; 2; 3; 4 |]     = [| [| 1; 2 |]; [| 2; 3 |]; [| 3; 4 |] |]
+
+  slide 3 [| 1; 2; 3; 4; 5 |]  = [| [| 1; 2; 3 |]; [| 2; 3; 4 |]; [| 3; 4; 5 |] |]
+
+  slide 5 [| 1; 2; 3; 4; 5 |]  = [| [| 1; 2; 3; 4; 5 |] |]
+*)
+
 
 
 module Cap =
--- a/src/batArray.mli
+++ b/src/batArray.mli
@@ -630,6 +630,16 @@ val equal : 'a BatOrd.eq -> 'a array BatOrd.eq
     equal if their lengths are the same and corresponding elements
     test equal. *)
 
+val slide : int -> 'a array -> 'a array array
+(** [slide n a] slides over [a] with windows of size [n].
+    [n] must be > 0 and [n] must be < [length [a]].
+    @raise Invalid_argument if [n] <= 0.
+    @raise Invalid_argument if [n] > [length [a]].
+    Each array in the result has size n.
+    Example: [slide 3 [| 1; 2; 3; 4; 5 |] = [| [| 1; 2; 3 |]; [| 2; 3; 4 |]; [| 3; 4; 5 |] |]]
+
+    @since NEXT_RELEASE *)
+
 (** {1 Override modules}*)
 
 (** The following modules replace functions defined in {!Array} with
--- a/src/batList.ml
+++ b/src/batList.ml
@@ -315,6 +315,27 @@ let ntake n l =
   ntake 2 [1; 2; 3; 4] = [[1; 2]; [3; 4]]
 *)
 
+let slide n l =
+  if n < 1 then invalid_arg "List.slide";
+  let rec loop dst l =
+    if length l < n then ()
+    else let taken = take n l in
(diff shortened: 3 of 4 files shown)


gasche:
One aspect that is giving me pause is the inefficiency of it, it is on O(M*N) in both time and memory due to the fact that you work on linked lists, whereas traversing the same elements with an iterator (for example) costs almost nothing, and is not necessarily harder to work with for the user.

Do you have an example use-case of this function, to see how the caller code looks like and whether a different approach to do the same result would avoid the fundamental inefficiency?

Sumandora:
> One aspect that is giving me pause is the inefficiency of it, it is on O(M*N) in both time and memory due to the fact that you work on linked lists, whereas traversing the same elements with an iterator (for example) costs almost nothing, and is not necessarily harder to work with for the user.

Fair enough, I'm somewhat new to OCaml and still need to get a grip on performance. Kinda forgot that we are working with linked lists in fact. There's probably a better way to write this, but I just wrote up something that seemed good enough with respect to the surrounding functions.

> Do you have an example use-case of this function, to see how the caller code looks like and whether a different approach to do the same result would avoid the fundamental inefficiency?

This is something I can help with however.
I was somewhat dissatisfied when doing a AOC puzzle, in particular year 2021, day 1, part 2.

I'm going to cut the mumbo jumbo and give you a straight explanation. You are given a list of numbers and you are being asked to sum them in triples and then say how many times the sums increase.
For example:
1,2,3,4,0 would become
1+2+3 -> 6
2+3+4 -> 7
6 < 7, so it inceases
2+3+4 -> 7
3+4+0 -> 7
7 = 7, so it doesn't increase.

I wrote it like this:
```ocaml
let part2 nums =
  let rec aux nums acc =
    match nums with
    | a :: b :: c :: d :: rest when a + b + c < b + c + d ->
      aux (b :: c :: d :: rest) (acc + 1)
    | _ :: rest -> aux rest acc
    | _ -> acc
  in
  aux nums 0
;;
```
This works fine, however it annoys me. The pattern match is ugly and I can't even change the amount nicely. Let's say the next question would be to do this with 100 elements instead of 3.

Now with List.slide, it becomes this:

```ocaml
let part2 nums =
  let sums = nums |> slide 3 |> List.map (List.fold ( + ) 0) in
  (* zipped could also be 'sums |> slide 2', however then we have a list of list and not a list of tuples, which is harder to exhaustively pattern match *)
  let zipped = List.combine (List.take (List.length sums - 1) sums) (List.tl sums) in
  zipped |> List.count_matching (Pair.fold ( < ))
;;
```

Not sure if this construction of `zipped` is optimal (edit: speaking of performance :p), but just to drive the point home, you could even say `sums |> List.slide 2` and get a pairwise iteration, although you wouldn't get these nice pairs, so your pattern match wouldn't be exhaustive. But still, I think it's lovely to see the "sliding-window-operation" pop up else where.

gasche:
For lists, there is no efficient way to add an element at the end of a list. So necessarily the operation of adding the next element to form the next sliding window is going to be O(W) (in both time and memory), in the width of the window, for a total complexity of O(N*W) (time and memory).

If you used arrays instead, you could avoid constructing the data structure using array access instead, and get an O(N*W) solution, but just O(1) memory.

In your problem, you don't actually need the sliding windows, but only the sum of their numbers. There is an efficient way to compute this, thanks to commutativity of integer addition: you can update the window by adding the next number and removing the previous number. This gives a solution in O(N) solution and O(1) memory.

Here are implementations that are parametric over the size of the sliding window (pick `3` for sums of triples):
```ocaml
(* array implementation *)
let part2 size nums =
  assert (size > 0);
  let nums = Array.of_list nums in
  let count = ref 0 in
  for i = size to Array.length nums - 1 do
    let remove = nums.(i - size) in
    let add = nums.(i) in
    if (add > remove) then incr count;
  done;
  !count

(* list implementation *)
let part2 size nums =
  let added = List.drop size nums in
  (* note: computed 'removed' is O(N) in memory, and not really necessary,
    but it makes the code look nicer (the lists have the same length so we can call fold_left2) *)
  let removed = List.take (List.length nums - size) nums in
  List.fold_left2 (fun acc add remove ->
    if add > remove then acc + 1 else acc
  ) 0 added removed
```

They are of course not general solutions to the problem you suggest, but they avoid the fundamental issue about `slide` being inefficient. It would be possible to propose a more abstract version that computes a fold on sliding windows, given functions to add and remove elements from the window result, but I wonder if the resulting complexity could be discouraging for users.

Sumandora:
> In your problem, you don't actually need the sliding windows, but only the sum of their numbers. There is an efficient way to compute this, thanks to commutativity of integer addition: you can update the window by adding the next number and removing the previous number. This gives a solution in O(N) solution and O(1) memory.

I'm aware, however for one, my solution is a much more literal translation of the problem, which is especially nice when reading it back, but on the other hand, this is not the only time I needed this:
https://github.com/Sumandora/sigscan-rs/blob/master/src/lib.rs
Here's a Rust library I made for finding byte patterns (needles) in haystacks.
The solution is to basically go over the haystack in windows (This is what Rust calls the sliding window) and getting the index of the first one that matches.
Although for big byte arrays, this would probably need to be adapted for sequences.

I'm just saying, that I don't exactly bother about the users performance, since when they use a function like this and there's a simpler solution then they are free to write the other one. But for users that don't bother about performance, this creates an easy way to encompass the problem into code.

UnixJunkie:
I never needed this one.
If another maintainer wants to include it, let us know.

Issue #1144, opened 2025-12-06
BatRandom.choice raise a private exception


rixed:
The mli file lacks the declaration.

To be honest I find the addition of this Empty exception was a bad idea.
The justification in #1080 was that it made the error less confusing and make Random.choice behave more like other modules.
I disagree on both points:

Random.choice on a empty enum is the poster child of "invalid argument". Do we believe that Invalid_arg should never be used at all and that we should have a distinct exception for every possible error path? Clearly, nobody really use exceptions like that, or I wouldn't be the first to notice this one not being catchable :)
Also, other modules that raise Empty are containers (queues, stack...), and they raise Empty when the container is empty, which makes perfect sense. But what does it mean for a random generator to be empty? Have it reached the end of the random sequence or what?
That's a bit confusing, while "invalid argument" points right at the culprit: the argument, which is invalid.

gasche:
If the exception is private, users can only catch it by writing a catch-all, and it means that the code they wrote is compatible with a change of exception back to `Invalid_argument`. So maybe you should not rush to make the exception public, and instead try to convince current Batteries maintainers (I'm not one of them anymore) to change the exception again?

(Note: maybe the problem was with raising `"Random.int"` which leaks an implementation detail; it is possible that raising `"Random.choice"` instead would be an improvement.)

rixed:
> maybe the problem was with raising "Random.int" which leaks an implementation detail; it is possible that raising "Random.choice" instead would be an improvement

You are right of course, and in the hope to spare everyone's time I've amended the commit.
Thank you for chiming in!

PR #1145, opened 2025-12-06
Fix Random.choice raising a private exception (#1144)

rixed:
See #1144

### DIFF
--- a/src/batRandom.ml
+++ b/src/batRandom.ml
@@ -21,7 +21,6 @@
  *)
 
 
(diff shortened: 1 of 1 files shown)


UnixJunkie:
I think changing the exception type is breaking compatibility.
So, I guess it should only be done if there is a new major release.
So, I would vote to close this PR.

UnixJunkie:
I tagged as for-next-major-release for the moment.

PR #1146, opened 2026-04-13
Add support for OCaml 5.5

kit-ty-kate:

PR #1147, opened 2026-04-25
Update GitHub Actions CI

sim642:
Should fix the inherent failures from #1146 and before.


UnixJunkie:
thanks for maintaining the CI

PR #1150, opened 2026-06-01
Add opam-repository-archive to CI for OCaml versions < 4.11

sim642:
The official opam-repository has recently gone through some archival of old OCaml compiler versions.
These could just be removed from the CI but I think using opam-repository-archive, which still contains them, is a better approach.

Since these versions are in the bounds that Batteries claims to support and given the legacy role of Batteries, I think it's worth having them in this CI, even if they won't be in opam-ci anymore.
This way it'll hopefully be more clear if some change breaks the support for very old versions and the decision to possibly drop their support in the future could be made intentionally, not accidentally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,11 +23,12 @@ jobs:
           - 4.13.x
           - 4.12.x
           - 4.11.x
-          - 4.10.x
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks!

